PR #32, opened 2021-10-11
Base template account views

IrgenKini:
Created Navigation Panel component  and integrated with Account preferences page. 

ClickUp: https://app.clickup.com/t/1kgg1k7

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -19,7 +19,6 @@ specifiers:
   '@tailwindcss/typography': ^0.4.1
   '@types/react': 17.0.24
   autoprefixer: 10.3.5
-  classnames: ^2.3.1
   clsx: 1.1.1
   editorjs-blocks-react-renderer: ^1.1.1
   eslint: 7.32.0
@@ -43,13 +42,12 @@ dependencies:
   '@apollo/client': 3.4.13_graphql@15.6.0+react@17.0.2
   '@headlessui/react': 1.4.1_react-dom@17.0.2+react@17.0.2
   '@heroicons/react': 1.0.4_react@17.0.2
-  '@saleor/sdk': 0.4.0-6_c8f63c1aa5501e06c5c457e6db1b8725
-  '@stripe/react-stripe-js': 1.6.0_02b0aafaeb8df084269af2860741c962
-  '@stripe/stripe-js': 1.19.1
-  '@tailwindcss/aspect-ratio': 0.2.2_tailwindcss@2.2.15
-  '@tailwindcss/forms': 0.3.4_tailwindcss@2.2.15
+  '@saleor/sdk': 0.4.0-7_c8f63c1aa5501e06c5c457e6db1b8725
+  '@stripe/react-stripe-js': 1.5.0_09a5e2cb3e4031e28165171142e9312f
+  '@stripe/stripe-js': 1.18.0
+  '@tailwindcss/aspect-ratio': 0.2.1_tailwindcss@2.2.15
+  '@tailwindcss/forms': 0.3.3_tailwindcss@2.2.15
   '@tailwindcss/typography': 0.4.1_tailwindcss@2.2.15
-  classnames: 2.3.1
   clsx: 1.1.1
   editorjs-blocks-react-renderer: 1.1.1
   graphql: 15.6.0
@@ -57,19 +55,19 @@ dependencies:
   next-seo: 4.28.1_6fbf59523517da779df50d355b375324
   react: 17.0.2
   react-dom: 17.0.2_react@17.0.2
-  react-hook-form: 7.17.1_react@17.0.2
+  react-hook-form: 7.15.4_react@17.0.2
   react-use: 17.3.1_react-dom@17.0.2+react@17.0.2
   sharp: 0.29.1
 
 devDependencies:
-  "@graphql-codegen/add": 3.1.0_graphql@15.6.0
-  "@graphql-codegen/cli": 2.2.0_graphql@15.6.0+typescript@4.4.3
-  "@graphql-codegen/introspection": 2.1.0_graphql@15.6.0
-  "@graphql-codegen/typescript": 2.2.2_graphql@15.6.0
-  "@graphql-codegen/typescript-apollo-client-helpers": 2.1.4_graphql@15.6.0
-  "@graphql-codegen/typescript-operations": 2.1.4_graphql@15.6.0
-  "@graphql-codegen/typescript-react-apollo": 3.1.4_graphql@15.6.0
-  "@types/react": 17.0.24
+  '@graphql-codegen/add': 3.1.0_graphql@15.6.0
+  '@graphql-codegen/cli': 2.2.0_graphql@15.6.0+typescript@4.4.3
+  '@graphql-codegen/introspection': 2.1.0_graphql@15.6.0
+  '@graphql-codegen/typescript': 2.2.2_graphql@15.6.0
+  '@graphql-codegen/typescript-apollo-client-helpers': 2.1.4_graphql@15.6.0
+  '@graphql-codegen/typescript-operations': 2.1.4_graphql@15.6.0
+  '@graphql-codegen/typescript-react-apollo': 3.1.4_graphql@15.6.0
+  '@types/react': 17.0.24
   autoprefixer: 10.3.5_postcss@8.3.7
   eslint: 7.32.0
   eslint-config-next: 11.1.2_82ad98da5e129ae9d310451a179848d9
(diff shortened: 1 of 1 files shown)


components/NavigationPanel.tsx
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable @next/next/link-passhref */
+import React from "react";
+import classNames from "classnames";
```
krzysztofwolski, line 3: We are already using clsx which serves the same purpose but is much faster lib.

components/NavigationPanel.tsx
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable @next/next/link-passhref */
```
krzysztofwolski, line 1: I would suggest wrapping components into `a` instead of ignoring the rule.

components/accountPreferences/PasswordPreferences.tsx
```diff
@@ -0,0 +1,108 @@
...
+              {...register("newPasswordRepeat", {
+                required: true,
+              })}
+            />
+            <p className="mt-2 text-sm text-red-600">{errors.error?.message}</p>
+          </div>
+        </div>
+        {successMessage !== null ? (
+          <p className="mt-2 text-sm text-green-600">{successMessage}</p>
+        ) : null}
```
krzysztofwolski, line 96:
```suggestion
        {!!successMessage &&
          <p className="mt-2 text-sm text-green-600">{successMessage}</p>
        }
```

package.json
```diff
@@ -14,12 +14,13 @@
...
     "@headlessui/react": "1.4.1",
     "@heroicons/react": "1.0.4",
+    "@saleor/sdk": "^0.4.0-6",
     "@stripe/react-stripe-js": "^1.5.0",
     "@stripe/stripe-js": "^1.18.0",
-    "@saleor/sdk": "^0.4.0-6",
     "@tailwindcss/aspect-ratio": "^0.2.1",
     "@tailwindcss/forms": "^0.3.3",
     "@tailwindcss/typography": "^0.4.1",
+    "classnames": "^2.3.1",
```
krzysztofwolski, line 23:
```suggestion

```
clsx is already here

pages/account/addressBook.tsx
```diff
@@ -0,0 +1,24 @@
+import BaseTemplate from "@/components/BaseTemplate";
+import { NavigationPanel } from "@/components/NavigationPanel";
+
+const AddressBookPage: React.VFC = () => {
+  return (
+    <BaseTemplate>
```
krzysztofwolski, line 6: I would like to have `AccountBaseTemplate` component, so this page would become:

```
<AccountBaseTemplate>
<div>Address Book Component goes here</div>
</AccountBaseTemplate>
```

Account pages have shared navigation so it would be a good idea to not reimplement layout at every page

components/accountPreferences/PasswordPreferences.tsx
```diff
@@ -0,0 +1,108 @@
+import { usePasswordChangeMutation } from "@/saleor/api";
+import React from "react";
+import { useForm } from "react-hook-form";
+
```
krzysztofwolski, line 4:
```suggestion
interface PasswordChangeFormData {
  oldPassword: string;
  newPassword: string;
  newPasswordRepeat: string;
}
```

If we create interface for the form, we can leverage type system later

components/accountPreferences/PasswordPreferences.tsx
```diff
@@ -0,0 +1,108 @@
...
+    null
+  );
+
+  const {
+    register,
+    handleSubmit,
+    formState: { errors },
+    setError,
+    clearErrors,
+  } = useForm();
```
krzysztofwolski, line 17:
```suggestion
  } = useForm<PasswordChangeFormData>();
```

components/accountPreferences/PasswordPreferences.tsx
```diff
@@ -0,0 +1,108 @@
...
+          oldPassword: formData.oldPassword,
+        },
+      });
+      const errors = result.data?.passwordChange?.errors || [];
+      if (errors.length > 0) {
+        errors.forEach((e) => setError("error", { message: e.message || "" }));
+        setTimeout(() => {
+          clearErrors();
+        }, 3000);
+        return;
```
krzysztofwolski, line 35:
```suggestion
errors.forEach((e) =>
        setError(e.field as keyof PasswordChangeFormData, {
          message: e.message || "",
        })
      );
      return;
```

components/accountPreferences/PasswordPreferences.tsx
```diff
@@ -0,0 +1,108 @@
...
+    register,
+    handleSubmit,
+    formState: { errors },
+    setError,
+    clearErrors,
+  } = useForm();
+
+  const onPasswordPreferenceSubmit = handleSubmit(async (formData) => {
+    if (formData.newPassword !== formData.newPasswordRepeat) {
+      setError("error", { message: "Passwords have to match." });
```
krzysztofwolski, line 21:
```suggestion
      setError("newPasswordRepeat", { message: "Passwords have to match." });
```

krzysztofwolski: Add errors to the relevant fields

components/accountPreferences/PasswordPreferences.tsx
krzysztofwolski: When we assign errors to the fields, error messages will be automatically hidden when the field is edited. Thats the default behavior of react-hook-form

components/accountPreferences/PasswordPreferences.tsx
```diff
@@ -0,0 +1,108 @@
...
+      });
+      const errors = result.data?.passwordChange?.errors || [];
+      if (errors.length > 0) {
+        errors.forEach((e) => setError("error", { message: e.message || "" }));
+        setTimeout(() => {
+          clearErrors();
+        }, 3000);
+        return;
+      } else if (result.data?.passwordChange?.user) {
+        setSuccessMessage("Password changed succesfully.");
```
krzysztofwolski, line 37:
```suggestion
        setSuccessMessage("Password changed successfully.");
```

components/accountPreferences/PasswordPreferences.tsx
```diff
@@ -0,0 +1,108 @@
...
+            </label>
+            <input
+              className="px-4 py-2 rounded-md text-sm outline-none w-full"
+              type="password"
+              placeholder="Old password"
+              {...register("oldPassword", {
+                required: true,
+              })}
+            />
+          </div>
```
krzysztofwolski, line 61:
```suggestion
            {!!errors.oldPassword && (
              <p className="mt-2 text-sm text-red-600">
                {errors.oldPassword.message}
              </p>
            )}
          </div>
```
And add errors to the rest of the fields accordingly.

components/accountPreferences/PasswordPreferences.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <p className="mt-2 text-sm text-red-600">{errors.error?.message}</p>
+          </div>
+        </div>
+        {successMessage !== null ? (
+          <p className="mt-2 text-sm text-green-600">{successMessage}</p>
+        ) : null}
+        <div>
+          <button
+            className="mt-2 w-40 bg-green-500 hover:bg-green-400 text-white py-2 rounded-md transition duration-100"
+            onClick={() => onPasswordPreferenceSubmit()}
```
krzysztofwolski, line 100:
```suggestion
            type="submit"
```

components/accountPreferences/EmailPreferences.tsx
```diff
@@ -0,0 +1,92 @@
+import React from "react";
```
krzysztofwolski, line 1: Please add error handling the same way as in the password change review comments.

krzysztofwolski (changes requested):

components/accountPreferences/EmailPreferences.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { useForm } from "react-hook-form";
+interface EmailChangeFormData {
+  newEmail: string;
+  password: string;
+  redirectUrl: string;
+}
+
+export const EmailPreferences: React.VFC<any> = ({}) => {
+  const [requestEmailChange] = useRequestEmailChangeMutation({});
+  const [successMessage, setSuccessMessage] = React.useState<String | null>(
```
zaiste, line 12: Can we relay here on the type inference to simplify this code? Is the `null` declaration necessary?

components/accountPreferences/EmailPreferences.tsx
```diff
@@ -0,0 +1,108 @@
...
+    formState: { errors },
+    setError,
+  } = useForm<EmailChangeFormData>();
+
+  const onEmailPreferenceSubmit = handleSubmit(async (formData) => {
+    const result = await requestEmailChange({
+      variables: {
+        newEmail: formData.newEmail,
+        password: formData.password,
+        redirectUrl: `https://localhost:3001/account`,
```
zaiste, line 27: Let's not hardcode URLs

components/Navbar.tsx
```diff
@@ -115,7 +115,7 @@ export const Navbar: React.VFC = ({}) => {
                       </p>
                     </div>
                     <div className="py-1">
-                      <Link href="/account">
+                      <Link href="/account/accountPreferences">
```
zaiste, line 118: Let's simplify a bit the path i.e. `/account/settings`

IrgenKini: it has that name because the user goes to a page where he can change his email and/or password. I can rename it for sure but I would have to rename also the component and "settings" may be a bit confusing for this page. What should I do ?

PR #38, opened 2021-10-19
Update README.md

culy247:


zaiste:
@culy247 thanks for the suggestion. The default package manager in this repo is `pnpm`, but `npm` should work without problem too.

PR #65, opened 2021-11-03
Search page

IrgenKini:


krzysztofwolski (changes requested):
One small bug:

- search for a "hoodie"
- open the hoodie product page
- click back in the browser

Products are displayed as expected, URL is ok, but search input does not contain search query

components/Navbar.tsx
```diff
@@ -45,7 +49,13 @@ export const Navbar = () => {
               </a>
             </Link>
           </div>
-          <div className="flex space-x-8">
+          <div className="flex space-x-px md:space-x-8 items-center">
+            <SearchIcon
+              className={
+                "flex-shrink-0 h-6 w-6 text-gray-400 hover:text-gray-500"
+              }
+              onClick={() => router.push("/search")}
```
krzysztofwolski, line 57: Please use `<Link>` component as rest of the icons in the navbar

pages/search.tsx
```diff
@@ -0,0 +1,38 @@
+import { useQueryState } from "next-usequerystate";
+import React from "react";
+import { useDebounce } from "react-use";
+
+import { BaseTemplate, ProductCollection } from "@/components";
+import { ProductFilterInput } from "@/saleor/api";
+
+const SearchPage = () => {
+  const [filter, setFilter] = useQueryState("search");
```
krzysztofwolski, line 9:
```suggestion
  const [filter, setFilter] = useQueryState("search", "");
```

I'm not familiar with this package - does it support default value? If so, we could skip `|| ""` at line 28

krzysztofwolski:
```suggestion
  const [searchQuery, setSearchQuery] = useQueryState("q");
```
IMO we can change the query argument to `q` - it's often used in other websites and makes the whole URL shorter.
I suggest changing filter to the `searchQuery` - it better describes contents of the variable (filter for products query can have more fields besides the search query)

pages/search.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <BaseTemplate>
+      <main className="max-w-7xl mx-auto w-full px-8 mt-5">
+        <p className="font-semibold text-2xl mb-5">Search</p>
+        <input
+          className="md:w-96 mb-10 block border-gray-300 rounded-md shadow-sm sm:text-sm"
+          type="text"
+          value={filter || ""}
+          placeholder="Search..."
+          onChange={(e) => setFilter(e.target.value)}
+        ></input>
```
krzysztofwolski, line 31:
```suggestion
        />
```

krzysztofwolski (changes requested):

krzysztofwolski (approved):

krzysztofwolski (approved):

PR #68, opened 2021-11-04
Product variant images

IrgenKini:


pages/products/[slug].tsx
```diff
@@ -130,6 +130,13 @@ const ProductPage = ({
 
   const productImage = product?.media![0];
 
+  let media;
```
Comment (krzysztofwolski) [inline pages/products/[slug].tsx:133]:
```suggestion
  const media = getGalleryMedia(product, selectedVariant); 
```

I would suggest creating a new function with docstring, which would explain which products are selected to be displayed, eg:
`If selected variants have media, return only those items. Otherwise, all product media are returned.`.

When we will return to this function a few months in the future, docstring will be very helpful :)

krzysztofwolski (changes requested):

krzysztofwolski (approved):

PR #77, opened 2021-11-09
Product attributes

IrgenKini:


pages/products/[slug].tsx
```diff
@@ -235,10 +268,55 @@ const ProductPage = ({
               >
                 {loadingAddToCheckout ? "Adding..." : "Add to cart"}
               </button>
-            ) : (
+            ) : null}
```
Comment (krzysztofwolski) [inline pages/products/[slug].tsx:271]: Line 262
```
{(selectedVariant && selectedVariant?.quantityAvailable > 0) && (
  <button
  ...
)}
```

No need for returning null :)

pages/products/[slug].tsx
```diff
@@ -235,10 +268,55 @@ const ProductPage = ({
...
+
+            {!selectedVariant && (
```
Comment (krzysztofwolski) [inline pages/products/[slug].tsx:273]: To avoid big shifts in the UI when selecting the variant, lets just disable "add to cart" button when variant is not selected

Comment (krzysztofwolski) [reply pages/products/[slug].tsx:273]: Button should be disabled also when variant is sold out. 

lets create new variable:
```
const isAddToCartButtonDisabled = !selectedVariant || selectedVariant?.quantityAvailable === 0
```

This way logic of disabling the button will be easier to follow and moved out of visual (rendering)

pages/products/[slug].tsx
```diff
@@ -235,10 +268,55 @@ const ProductPage = ({
...
+                <RichText jsonStringData={product.description} />
+              </div>
+            )}
+
+            {attributes.length > 0 && (
+              <div>
+                <p className="text-lg mt-2 font-medium text-gray-500">
+                  Attributes
+                </p>
+                <div className="">
```
Comment (krzysztofwolski) [inline pages/products/[slug].tsx:296]:
```suggestion
                <div className="">
```
If there is no classes, attribute should be omitted.

pages/products/[slug].tsx
```diff
@@ -235,10 +268,55 @@ const ProductPage = ({
...
+                  {attributes.map((attribute: any) => (
+                    <div
+                      key={attribute.attribute.name}
+                      className="grid grid-cols-2"
+                    >
+                      <div>
+                        <p>{attribute.attribute.name}</p>
+                      </div>
+                      <div>
+                        {attribute.values.map((value: any, index: number) => (
```
Comment (krzysztofwolski) [inline pages/products/[slug].tsx:306]:
```suggestion
                        {attribute.values.map((value, index) => (
```
TS should infer types of attributes.

krzysztofwolski (changes requested):

krzysztofwolski (approved):

PR #79, opened 2021-11-09
Saved addresses in checkout

IrgenKini:


components/checkout/ChoosePredefinedAddress.tsx
```diff
@@ -0,0 +1,76 @@
...
+  useUserAddressesQuery,
+} from "@/saleor/api";
+
+import { AddressFormData } from "./AddressForm";
+
+interface ChoosePredefinedAddressProps {
+  updateAddressMutation: (address: AddressFormData) => Promise<CheckoutError[]>;
+}
+
+export const ChoosePredefinedAddress = ({
```
krzysztofwolski, line 18: I would rename it to the `SavedAddressSelect`.

@zaiste any suggestions?

components/checkout/ChoosePredefinedAddress.tsx
```diff
@@ -0,0 +1,76 @@
...
+      country: CountryCode.Pl,
+      streetAddress1: address.streetAddress1,
+      city: address.city,
+      postalCode: address.postalCode,
+    });
+  };
+
+  return (
+    <div className="grid grid-cols-2 mb-2">
+      {addresses.length > 0 &&
```
krzysztofwolski, line 48: If there is no address in the list, whole component should return null instead of empty div

krzysztofwolski (changes requested):

components/checkout/ChoosePredefinedAddress.tsx
krzysztofwolski: @IrgenKini Contacted with Zaiste and we settled for SavedAddressSelectionList

components/checkout/ShippingAddressSection.tsx
```diff
@@ -83,6 +85,11 @@ export const ShippingAddressSection = ({
         <>
           {editing ? (
             <>
+              <SavedAddressSelectionList
```
krzysztofwolski, line 88: Render component only if a user is authenticated (ATM code fire API call even if checkout is anonymous)

components/checkout/SavedAddressSelectionList.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const [selectedSavedAddres, setSelectedSavedAddress] =
+    React.useState<AddressDetailsFragment | null>();
+
+  if (loading) {
+    return <Spinner />;
+  }
+
+  if (error) return <p>Error : {error.message}</p>;
+
+  let addresses = data?.me?.addresses || [];
```
krzysztofwolski, line 31:
```suggestion
  const addresses = data?.me?.addresses || [];
  if (addresses.length === 0) return null;
  return (
    <div className="grid grid-cols-2 mb-2">
    ...
```

I would suggest trying to use the same pattern as the rest of the repo.

krzysztofwolski (changes requested):

krzysztofwolski (approved):

PR #81, opened 2021-11-09
Split codegen

krzysztofwolski:
Wanted to explore GQL Code Generator feature for splitting the output code.


krzysztofwolski:
Discussion:
I quite liked how imports worked before (import from `.api` seemed nice). Big pro IMO is `__generated__` in the path - I think it's more clear for the newcomers that there's a tool that handles all the heavy lifting.

Imports example:
```ts
// old import
import { useMainMenuQuery } from "@/saleor/api";
// "new", splitted import
import { useMainMenuQuery } from "graphql/queries/__generated__/MainMenu";

export const MainMenu = () => {
  const { loading, error, data } = useMainMenuQuery(); // usage is exactly the same
```

krzysztofwolski:
For now, we decided to leave this idea for the future. DX impact is not obvious and we see a more complicated config as a con.

We might return to this idea later on.

PR #94, opened 2021-12-13
Channels selection dialog

bmigirl:
What's done:
- Channels selection dialog
- Channels provider that saves current channel to local storage (possibly to be removed, depending on the later implementation - leave be for now)
- Channels config file
- `usePaths` hook that returns single value { paths } to be used as `paths.products(slug, queryParams)` (to add more paths add values to the `basicPaths` in the same file as `usePaths`

What needs to be done:
- change all currently hardcoded paths to use `usePaths` (as above)
- entities such as products, collections, categories should be fetched generated at server build for generating static props and paths (started work on this in `components/utils/serverUtils.ts` but it's not tested yet)
- typing


zaiste:
@bmigirl great progress!

(minor) could you change the component declarations from `const A: React.FC<T>` to `const A = ({...}: T) =>`? we follow [those guidelines](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/)

krzysztofwolski:
TODO:
- sitemap with channels and locales
- redirection defined @ config.next.js should be using defaults from the lib/regions
- locales should be defined as objects, same as channels. Ex `{slug: "en-US", name: "American English"}`

lib/graphql.ts
```diff
@@ -17,7 +17,7 @@ const apolloClient = new ApolloClient({
 
 export const saleorClient = createSaleorClient({
   apiUrl: API_URI,
-  channel: DEFAULT_CHANNEL,
+  channel: "default-channel",
```
krzysztofwolski, line 20: Should be imported from regions.ts

lib/paths.ts
```diff
@@ -0,0 +1,10 @@
+import { useRouter } from "next/router";
+
+import { pagesPath } from "@/lib/$path";
+
+export const usePaths = () => {
+  const { query } = useRouter();
+  const channel = query.channel?.toString() || "default-channel";
```
krzysztofwolski, line 7: Defaults should be imported from regions

lib/typePolicies.ts
```diff
@@ -8,7 +8,7 @@ export const typePolicies: TypedTypePolicies = {
   Money: {
     fields: {
       localizedAmount: {
-        read(_, { readField }) {
+        read(_, { readField, args }) {
```
krzysztofwolski, line 11: TODO: format should take locale to format the price.
It might be a show stopper for extending local schema, because not all queries will need locale query argument (needs to be validated)

package.json
```diff
@@ -34,6 +35,8 @@
     "next-seo": "^4.28.1",
     "next-sitemap": "^1.6.203",
     "next-usequerystate": "1.4.0-beta.5",
+    "pathpida": "^0.17.0",
+    "qs": "^6.10.1",
```
krzysztofwolski, line 39: Can be removed

bmigirl:
As discussed with @krzysztofwolski the sitemap will be added in a separate pr, also the locales as object will be (already done [here](https://github.com/saleor/react-storefront/pull/101). Rest of the comments has been addressed

components/RegionDialog/RegionDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+}
+
+const RegionDialog = ({ isOpen, onClose }: RegionDialogProps) => {
+  const router = useRouter();
+  const { channels, currentChannel, setCurrentChannel } = useChannels();
+
+  const { register, handleSubmit } = useForm<RegionFormData>({
+    defaultValues: {
+      channel: currentChannel.slug,
+      locale: router.locale || DEFAULT_LOCALE,
```
jwm0, line 28: should `router.locale` ever be null or undefined?

components/RegionDialog/RegionDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  const onSubmit = handleSubmit(async (formData: RegionFormData) => {
+    if (formData.channel !== currentChannel.slug) {
+      await setCurrentChannel(formData.channel);
+      clearCheckout();
+    }
+    onClose();
+
+    // Update current URL to use the chosen channel
+    await router.push({
```
jwm0, line 40: `await` not needed

components/RegionDialog/RegionDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+      },
+    });
+  });
+
+  if (!isOpen) {
+    return null;
+  }
+
+  return (
+    <div
```
jwm0, line 55: Modals should always be wrapped inside React Portals, perhaps it'd be a good idea to use a headlessui from taildwind? https://headlessui.dev/react/dialog

components/SaleorProviderWithChannels/SaleorProviderWithChannels.tsx
```diff
@@ -0,0 +1,23 @@
+import { SaleorProvider } from "@saleor/sdk";
+import React, { useEffect } from "react";
+
+import { saleorClient } from "@/lib/graphql";
```
jwm0, line 4: Bump @saleor/sdk to latest version (0.4.1), it exports a `useSaleorConfig` hook that has a `setChannel` method. Use that instead of importing the client itself

lib/paths.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { pagesPath } from "@/lib/$path";
+
+import { DEFAULT_CHANNEL,DEFAULT_LOCALE } from "./regions";
+
+export const usePaths = () => {
+  const { query } = useRouter();
+  const channel = query.channel?.toString() || DEFAULT_CHANNEL.slug;
+  const locale = query.locale?.toString() || DEFAULT_LOCALE;
+  return { paths: pagesPath._channel(channel)._locale(locale) };
```
jwm0, line 11: Why is it artificially nested inside an object? Wouldn't this usage make more sense?
```
const paths = usePaths();
```

lib/ssr/category.ts
```diff
@@ -0,0 +1,61 @@
...
+          },
+        }
+      );
+
+    const edges = response.data.categories?.edges;
+    if (!edges) {
+      break;
+    }
+    const responseSlugs: string[] = edges.map((edge) => edge.node.slug);
+    for (let channel of CHANNELS) {
```
jwm0, line 42: Why is for loop mixed with forEach? Let's choose a single iterating approach and stick to it:
```
    CHANNELS.forEach((channel) => {
      const channelSlug = channel.slug;

      LOCALES.forEach((locale) => {
        responseSlugs.forEach((slug) => {
          paths.push({
            params: {
              channel: channelSlug,
              locale: locale,
              slug,
            },
          });
        });
      });
    });
```

lib/ssr/homepage.ts
```diff
@@ -0,0 +1,21 @@
...
+  const paths: Path<ProductPathArguments>[] = [];
+  regionCombinations().forEach((combination) => {
+    paths.push({
+      params: {
+        locale: combination.localeSlug,
+        channel: combination.channelSlug,
+      },
+    });
+  });
+  return paths;
```
jwm0, line 20:
```suggestion
  const paths: Path<ProductPathArguments>[]  = regionCombinations().map((combination) => 
  ({
      params: {
        locale: combination.localeSlug,
        channel: combination.channelSlug,
      },
    });
  );
  return paths;
```

pages/[channel]/[locale]/category/[slug].tsx
```diff
@@ -62,22 +42,32 @@ const CategoryPage = ({
...
   return {
     paths: paths,
     fallback: true,
   };
-}
+};
+
+export const getStaticProps = async (context: GetStaticPropsContext) => {
+  const categorySlug = context.params?.slug?.toString() as string;
+  const locale = context.params?.locale?.toString() as string;
```
Comment (jwm0) [inline pages/[channel]/[locale]/category/[slug].tsx:55]: Unused property

pages/[channel]/[locale]/category/[slug].tsx
```diff
@@ -62,22 +42,32 @@ const CategoryPage = ({
...
+  const paths = await categoryPaths();
   return {
     paths: paths,
     fallback: true,
   };
-}
+};
+
+export const getStaticProps = async (context: GetStaticPropsContext) => {
+  const categorySlug = context.params?.slug?.toString() as string;
```
Comment (jwm0) [inline pages/[channel]/[locale]/category/[slug].tsx:54]:
```suggestion
  const categorySlug = context.params?.slug?.toString()!;
```
Assert the type instead

pages/[channel]/[locale]/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <>
+      <BaseSeo />
+      <div className="py-10">
+        <header className="mb-4">
+          <div className="max-w-7xl mx-auto px-8"></div>
+        </header>
+        <main>
+          <div className="max-w-7xl mx-auto px-8">
+            {menuData?.menu?.items?.map((m) => {
+              if (!!m) return <HomepageBlock key={m?.id} menuItem={m} />;
```
Comment (jwm0) [inline pages/[channel]/[locale]/index.tsx:26]:
```suggestion
              if (m) return <HomepageBlock key={m?.id} menuItem={m} />;
```
value is already cast to boolean inside condition

storefront.config.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  defaultLocale: "en-US",
+  locales: ["fr-fr", "pl-pl"],
```
jwm0, line 3: `locales` should contain a list of all possible languages including the default one.

storefront.config.js
```diff
@@ -0,0 +1,21 @@
...
+  defaultChannel: {
+    slug: "default-channel",
+    name: "United States Dollar",
+    currencyCode: "USD",
+  },
+  channels: [
```
jwm0, line 9: Same goes for channels, there should be a single source for all existing channels. The `defaultChannel` could use a `slug` instead.

jwm0 (changes requested):

components/RegionDialog/RegionDialog.tsx
krzysztofwolski: Had strange behavior without it, please test the change. Not sure if it was a coincidence or a real issue.

components/RegionDialog/RegionDialog.tsx
bmigirl: I think since css is here is not to show best practices but only to make the example easy to navigate etc. I'd just stick to the simplest solution we can get and leave it just with portal

components/SaleorProviderWithChannels/SaleorProviderWithChannels.tsx
krzysztofwolski: Since this hook is not inside the Saleor provider, I cannot use the hook. 
Will back to it in separate PR.

pages/[channel]/[locale]/index.tsx
Comment (krzysztofwolski) [reply pages/[channel]/[locale]/index.tsx:26]: Used pattern existing in the rest of the project. Leaving it to be resolved in separate PR.

components/ChannelsProvider/ChannelsProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+  currentChannel: Channel;
+  setCurrentChannel: (slug: string) => void;
+}
+
+export const ChannelsContext = createContext<ChannelsConsumerProps>({
+  channels: CHANNELS,
+  defaultChannel: DEFAULT_CHANNEL,
+  currentChannel: DEFAULT_CHANNEL,
+  setCurrentChannel: () => {},
+});
```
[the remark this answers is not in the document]
krzysztofwolski: Implemented.

components/RegionDialog/RegionDialog.tsx
krzysztofwolski: When user navigates to the rootpath without the locale/channel

krzysztofwolski:
Reverted for now config file:
- had issues with adding types to the config
- wanted to avoid additional issues with merging translations PR

Config file proposal will be submitted as separate PR

storefront.config.js
krzysztofwolski: Storefront config file will be submitted as separate PR.

lib/ssr/category.ts
```diff
@@ -0,0 +1,61 @@
...
+      let channelSlug = channel.slug;
+      for (let locale of LOCALES) {
+        responseSlugs.forEach((slug) => {
+          paths.push({
```
zaiste, line 46: This is O(n^3)

zaiste (approved):

PR #97, opened 2021-12-15
Refactor product page

krzysztofwolski:
Extracted parts of the product page into separate components to make the page easier to work with :)
Also changed URL of product details to singular form to make it consistent with the rest of the project.


lib/media.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const getYouTubeIDFromURL = (url: string) => {
+  var regExp =
+    /^.*((youtu.be\/)|(v\/)|(\/u\/\w\/)|(embed\/)|(watch\?))\??v?=?([^#&?]*).*/;
+  var match = url.match(regExp);
+  return match && match[7].length == 11 ? match[7] : undefined;
+};
+
+export const getVideoThumbnail = (videoUrl: string) => {
+  return `https://img.youtube.com/vi/${getYouTubeIDFromURL(
```
zaiste, line 34: @krzysztofwolski you could use an _implicit_ `return` here

lib/media.ts
```diff
@@ -0,0 +1,37 @@
...
+  product: ProductDetailsFragment;
+  selectedVariant?: ProductVariantDetailsFragment;
+}) => {
+  if (selectedVariant && selectedVariant.media?.length !== 0)
+    return selectedVariant.media?.filter(notNullable) || [];
+  return product?.media?.filter(notNullable) || [];
+};
+
+export const getYouTubeIDFromURL = (url: string) => {
+  var regExp =
```
zaiste, line 27: @krzysztofwolski how about we stick to `const` / `let`?

zaiste (approved):
@krzysztofwolski awesome! :)

PR #102, opened 2021-12-21
Remove layout component

bmigirl:
As per docs, since we have single layout for all our pages, we don't need to set it dynamically per component. Hence:
- Remove get layout usages 
- Move Layout component to _app.tsx

Needs merging #94 first


zaiste:
@bmigirl could you cherry pick the last commit (ee3b1d82840200310aa08c5d0a067d9f0c1abbd1) and issue a new PR?

PR #112, opened 2022-01-04
Enums-as-types

krzysztofwolski:
Client bundle size before: 731 KB
after:  693 KB
(gzipped 219 vs 206)

### DIFF
--- a/codegen.yml
+++ b/codegen.yml
@@ -17,8 +17,8 @@ generates:
     config:
       scalars:
         JSONString: string
-        # SampleScalar: Scalar.SampleScalar
       dedupeOperationSuffix: true # Prevent suffix duplication in generated names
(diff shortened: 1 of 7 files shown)


zaiste:
@krzysztofwolski it's similiar to my previous findings as suggested by The Guild with `onlyOperationTypes: true`. It decreases the bundle by 5-10%

jwm0:
@zaiste I'd argue that this provides a better DX as it no longer requires additional import which can be especially annoying if you extract the functions elsewhere. Having them as string unions gives us immediate suggestions and provides the exact same type-safety and makes it more resilient to changes in enum names on the Saleor side (as long as the value remains the same, no code changes are necessary)

zaiste:
@jwm0 You're right. Your comment changed my mind.

PR #114, opened 2022-01-11
Fix fresh checkout cart amount

bmigirl:
Currently when checkout is null and we add the first item, the cart icon in navbar still displays 0 before some other rerenders. It's because our currently used `useLocalStorage` [doesn't listen to changes in the storage](https://github.com/streamich/react-use/issues/2269). Hence when checkout token is changed in the background, it doesn't update in the navbar right away most of the times. Added local implementation to lib/hooks with useLocalStorage taken directly from our cloud repo. Didn't change all the other usages because I think we might want to decide when we actually want to listen to the changes + maybe react-use will get this added in the future.


krzysztofwolski (changes requested):
Tested this on incognito mode and counter does not update :/

bmigirl:
- Added checkout by token provider to keep all that data centered (also fixes problem with cart not update'ing)
- Kept the local lib version of useLocalStorage for syncing between tabs when changes are done to local storage
- Removed some prop passes of checkout (left those where it is given that checkout will exist to avoid unnecessary conditionals)

krzysztofwolski (changes requested):

krzysztofwolski (changes requested):
After creating the order, application persist old token of the closed checkout and prevents user from adding new products

krzysztofwolski (approved):

PR #130, opened 2022-01-26
Use urql

zaiste:
It's mostly about the size in the client bundle i.e. `7kb` (urql) vs `32kb` (apollo)


patrys:
Task linked: [SALEOR-5628 Evaluate URQl](https://app.clickup.com/t/2549495/SALEOR-5628)

PR #147, opened 2022-02-17
On-demand invalidation

krzysztofwolski:


package.json
```diff
@@ -35,12 +35,12 @@
...
     "graphql": "16.3.0",
-    "next": "12.0.10",
+    "next": "12.1.0",
     "next-seo": "^5.1.0",
-    "next-sitemap": "^2.1.13",
+    "next-sitemap": "^2.4.2",
     "next-usequerystate": "1.7.0",
-    "react": "rc",
-    "react-dom": "rc",
+    "react": "17.0.2",
```
zaiste, line 42: @krzysztofwolski let's keep React 18 RC to test and simplify the migration when it's released

zaiste (changes requested):

package.json
krzysztofwolski: Had to reverse to 17 because of problems with the new version of Next

Qasem-h (approved):

krzysztofwolski:
Before we introduce in-demand invalidation, we'll need an additional script for automatically subscribing to the events. Closing for now.

PR #156, opened 2022-03-18
Configurable homepage, changed variant selector, bugfixes

krzysztofwolski:
Changes:
- variant selector as a select widget
- payment details prepopulated for the DEMO mode
- changed links after order completed
- fixed layout for the account pages
- image optimization will allow for the images from API endpoint out of the box (instead of hardcoded value)
- redirection out of the checkout pages if the cart is empty

### DIFF
--- a/components/checkout/payments/DummyCreditCardSection.tsx
+++ b/components/checkout/payments/DummyCreditCardSection.tsx
@@ -5,6 +5,7 @@ import { useIntl } from "react-intl";
 
 import { useRegions } from "@/components/RegionsProvider";
 import { messages } from "@/components/translations";
+import { DEMO_MODE } from "@/lib/const";
 import { usePaths } from "@/lib/paths";
 import { useCheckout } from "@/lib/providers/CheckoutProvider";
 import {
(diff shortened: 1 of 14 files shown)


Qasem-h (approved):

components/checkout/payments/DummyCreditCardSection.tsx
```diff
@@ -43,17 +44,25 @@ export const DummyCreditCardSection = ({
     total: formatPrice(totalPrice),
   });
 
+  let defaultValues = DEMO_MODE
```
jwm0, line 47:
```suggestion
  const defaultValues = DEMO_MODE
```
I don't see them being reassigned

components/product/VariantSelector.tsx
```diff
@@ -17,40 +16,47 @@ export const VariantSelector = ({
...
 
   const variants = product.variants;
+
+  // Skip displaying selector when theres less than 2 variants
   if (!variants || variants.length === 1) {
     return null;
   }
+
+  const onChange = (event: React.FormEvent<HTMLSelectElement>) => {
+    const target = event.target as HTMLSelectElement;
```
jwm0, line 31:
```suggestion
    const target = event.target;
```
This type should be auto-inferred based on the fn argument, are you sure it's the right type?

krzysztofwolski: StackOverflow for the rescue - changed to `React.ChangeEvent<HTMLSelectElement>` which seems to be the right one.

jwm0 (approved):

PR #160, opened 2022-03-24
Apply airbnb code style

krzysztofwolski:
Changes:
- Add commit hooks for the linters (only the staged files are checked)
- Split CI lint and build into separate jobs
- Adjusting code style to the new preset (airbnb)
- A few type corrections 

Rules which require discussion:
- https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/function-component-definition.md - It's on by default, should we keep it?
- turned off rules in the eslint config in the discussion section: https://github.com/saleor/react-storefront/pull/160/files#diff-e7e195d3d8fa5b82fa51ed952f3a179d429cbc46c0432ce1f7b357c54c162822


Qasem-h (approved):

adrian-potepa (approved):

jwm0 (approved):
>https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/function-component-definition.md - It's on by default, should we keep it?

I like it, component names will show up in profiler by default without having to use `.displayName = ...`

PR #166, opened 2022-04-01
Provide default values in ENVs for simplicity

zaiste:
The current `.env` is a source of confusion (e.g. [this question](https://github.com/saleor/react-storefront/discussions/164)). Let's set the _default_ values in the code whenever and use environment only to overwrite that or set something that cannot have a default value (i.e. a GraphQL endpoint).

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,1 @@
 NEXT_PUBLIC_API_URI=https://vercel.saleor.cloud/graphql/
-NEXT_PUBLIC_DEFAULT_CHANNEL=default-channel
-NEXT_PUBLIC_HOMEPAGE_MENU=homepage
-NEXT_PUBLIC_VERCEL_URL=https://localhost:3001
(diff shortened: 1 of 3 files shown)


krzysztofwolski:
Don't think it's simplification - having all variables in the `.env` file IMO is easier to follow than looking thru the repository and looking where we are using the variables. Not every variable can be pulled from the const.ts file too - for example yaml config for the codegen.

This question from the community should be resolved via documentation of the project.

krzysztofwolski (changes requested):
Don't think it's simplification - having all variables in the `.env` file IMO is easier to follow than looking thru the repository and looking where we are using the variables. Not every variable can be pulled from the const.ts file too - for example yaml config for the codegen.

This question from the community should be resolved via documentation of the project.

krzysztofwolski:
Docs pages added in the separate PR #167

Qasem-h (approved):

PR #170, opened 2022-04-04
Add geolocation

krzysztofwolski:
Removed rigid region redirection from `next.config.js` file in favor of middleware.

By default, geolocation is turned off, because redirecting to the proper channel is depending on shop configuration. I think it's a nice starting point for implementing production storefronts.

This implementation only recognizes pln channel based on user country. Locale is chosen from existing languages, based on browser preferences.

Please note - geolocation does not work on local machines.


krzysztofwolski:
Turned geolocation on for this PR preview deployment

patrys:
Task linked: [SALEOR-4763 Add language redirection when entering site from different locale](https://app.clickup.com/t/2549495/SALEOR-4763)

adrian-potepa (approved):

Qasem-h (approved):

pages/_middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+import { GEOLOCATION } from "@/lib/const";
+import { CHANNELS, DEFAULT_CHANNEL, DEFAULT_LOCALE, LOCALES } from "@/lib/regions";
+
+export function LocaleRedirectionMiddleware({
+  nextUrl,
+  headers,
+  geo,
+}: NextRequest): NextMiddlewareResult | Promise<NextMiddlewareResult> {
+  if (nextUrl.pathname !== "/") {
+    // redirect should only be applied on homepage, without any region/locale chosen
```
jwm0, line 13: I think this is too narrow, it should catch all routes that don't include region/locale. If you remove locale/region from products pages it will throw an error

krzysztofwolski: By error you mean 404? That's expected. 
I would like to have geolocation only for the homepage. If the user visits the storefront from page URL IMO we should **not** redirect automatically. 
If there's a difference between URL and geolocation we could show a notification, but that's outside of the scope of this PR.

jwm0 (approved):

PR #171, opened 2022-04-05
Use headless component for the modal

krzysztofwolski:
- clicking outside of the modal closes it. Same for the hitting the escape button
- headless UI provides additional aria args

### DIFF
--- a/components/RegionDialog/RegionDialog.tsx
+++ b/components/RegionDialog/RegionDialog.tsx
@@ -1,3 +1,4 @@
+import { Dialog } from "@headlessui/react";
 import { useRouter } from "next/router";
 import React from "react";
 import { useForm } from "react-hook-form";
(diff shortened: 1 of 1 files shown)


components/RegionDialog/RegionDialog.tsx
```diff
@@ -54,56 +55,73 @@ export function RegionDialog({ isOpen, onClose }: RegionDialogProps) {
...
-            ))}
-          </select>
-        </div>
-        <div className="p-3  mt-4 mb-4 text-center space-x-4 md:block">
-          <Button onClick={onSubmit}>{t.formatMessage(messages.saveButton)}</Button>
+    <Dialog as="div" className="fixed inset-0 z-10 overflow-y-auto" open={isOpen} onClose={onClose}>
+      <div className="min-h-screen px-4 text-center">
+        <Dialog.Overlay className="fixed inset-0 bg-black opacity-50" />
+
+        {/* This element is to trick the browser into centering the modal contents. */}
```
jwm0, line 62: It looks like this can be achieved with just pure css?

krzysztofwolski: Yup. But decided that don't want to introduce another approach before the redesign and went with the pure tailwind.

jwm0: By pure css I mean using tailwind and not using html tricks to center contents. This should be easily done with flex

krzysztofwolski: You are totally right (alignment pun intended)

jwm0 (approved):

PR #179, opened 2022-04-14
Saleor 6298/footer tailwind theme

adrian-potepa:


components/Footer/Footer.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const { data, error } = useFooterMenuQuery({ variables: { ...query } });
+
+  if (error) {
+    console.error("Footer component error", error.message);
+    return null;
+  }
+
+  const menu = data?.menu?.items || [];
+
+  const gethLinkPath = (item: MenuItemFragment) => {
```
krzysztofwolski, line 28: WDYT about moving this function to the `lib/menus.ts`, it seems like the main menu would use the same code?

pages/_document.tsx
```diff
@@ -6,7 +6,7 @@ class MyDocument extends Document {
     const { hostname } = new URL(uri);
 
     return (
-      <Html lang="en" className="bg-gray-50">
+      <Html lang="en">
```
krzysztofwolski, line 9: Shouldn't language be taken from the Regions provider?

krzysztofwolski, line 9: Shouldn't language be taken from the Regions provider?

styles/globals.css
```diff
@@ -23,7 +25,7 @@
   }
 
   a {
-    @apply text-blue-600 hover:text-blue-500 cursor-pointer;
+    /* @apply text-blue-600 hover:text-main cursor-pointer; */
```
krzysztofwolski, line 28: Comments can be dropped

krzysztofwolski, line 28: Comments can be dropped

krzysztofwolski (changes requested):
Besides comments LGTM

pages/_document.tsx
adrian-potepa: Fixed:

```
// _document.tsx

class MyDocument extends Document<{ lang?: string }> {
  static async getInitialProps(ctx: DocumentContext) {
    const initialProps = await Document.getInitialProps(ctx);

    return { ...initialProps, lang: ctx?.query?.locale };
  }

  render() {
    const uri = process.env.NEXT_PUBLIC_API_URI!;
    const { hostname } = new URL(uri);

    return (
      <Html lang={this.props.lang}>
        <Head>
          <link rel="preconnect" href={`//${hostname}`} crossOrigin="true" />
          <link rel="dns-prefetch" href={`//${hostname}`} />
        </Head>
        <body>
          <Main />
          <NextScript />
        </body>
      </Html>
    );
  }
}

export default MyDocument;

```

components/Footer/Footer.tsx
adrian-potepa: Done.

krzysztofwolski (approved):

components/Footer/Footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+export function Footer({ className, ...rest }: FooterProps) {
+  const paths = usePaths();
+  const { query, currentChannel, currentLocale } = useRegions();
+
+  const { data, error } = useFooterMenuQuery({ variables: { ...query } });
+
+  if (error) {
+    console.error("Footer component error", error.message);
+    return null;
+  }
```
jwm0, line 25: I'd just print the error to console but not return early, since there are already guards against `data` not being defined

components/RegionsProvider/RegionsProvider.tsx
```diff
@@ -81,6 +82,7 @@ export function RegionsProvider({ children }: PropsWithChildren<{}>) {
 
   return (
     <Provider value={providerValues}>
+      <Head>{/* <html lang={providerValues.currentLocale} /> */}</Head>
```
jwm0, line 85: unused

adrian-potepa: Removed.

components/Footer/Footer.tsx
adrian-potepa: Removed line with null return.

PR #182, opened 2022-04-21
Saleor 6620/new main menu

adrian-potepa:
New navbar design.

Figma: https://www.figma.com/file/G6DxCJtgsC1D5bLM0WZozt/?node-id=6143%3A5782


components/BurgerMenu/BurgerMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+    variables: { ...query },
+  });
+
+  if (error) {
+    console.error("BurgerMenu component error", error.message);
+  }
+
+  const onLogout = async () => {
+    // clear all the user data on logout
+    await logout();
```
bmigirl, line 44: Since this logic is used in multiple places, it should be a hook e.g. 
```
const { logout } = useLogout();

File: components/BurgerMenu/BurgerMenu.tsx [hunk 2]
```diff
@@ -0,0 +1,103 @@
...
+        [styles["container--open"]]: open,
+      })}
+    >
+      <div className={styles.backdrop} aria-hidden="true" onClick={onCloseClick} />
+      <div className={styles.body}>
+        <div className="flex justify-end w-full mb-5">
+          <NavIconButton icon="close" onClick={onCloseClick} />
+        </div>
+        {menu.map((item) => {
+          if (!item) return null;
```
bmigirl, line 64: What's the case in the api for menu item being null or undefined? It looks weird and maybe it's a change we should make in the api then to not return null

components/BurgerMenu/CollapseMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+import { useRegions } from "../RegionsProvider";
+import styles from "./BurgerMenu.module.css";
+import ChebronDown from "./chevronDownIcon.svg"; // in the final version it should be imported from ui-kit package
+
+interface CollapseMenuProps {
+  data: MenuItemWithChildrenFragment;
+}
+
+function SubCollapseMenu({ data }: CollapseMenuProps) {
```
bmigirl, line 17: Separate files please, it's always difficult to navigate when in some files there are single components and in some other many

components/BurgerMenu/CollapseMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+    <Link href={getLinkPath(data!, currentChannel.slug, currentLocale)} passHref>
+      <a href="pass" className={styles["collapse-sub"]}>
+        {translate(data, "name")}
+      </a>
+    </Link>
+  );
+
+  return (
+    <div className="mt-4">
+      {!data?.children?.length ? (
```
bmigirl, line 35: Just to prittify a lil bit
```
const shouldDisplayAnchor = !data?.children?.length
```

components/BurgerMenu/CollapseMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+interface CollapseMenuProps {
+  data: MenuItemWithChildrenFragment;
+}
+
+function SubCollapseMenu({ data }: CollapseMenuProps) {
+  const [open, setOpen] = useState(false);
+  const { currentChannel, currentLocale } = useRegions();
+
+  const anchor = data.url ? (
```
bmigirl, line 21: Since this is practically the same (except classnames) in both of these component, it should probably be a component as well, e.g `MenuAnchor` or whatever. Then we could just 
```
 <div className="mt-4">
      {!data?.children?.length ? (
          <MenuAnchor
```

components/Navbar/DropdownMenu.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+function Dropdown({ data }: DropdownProps) {
+  const {
+    currentChannel: { slug },
+    currentLocale,
+  } = useRegions();
+
+  return (
+    <div className={styles.dropdown}>
+      {data?.url ? (
```
bmigirl, line 22: Same as above, could be moved to a component

components/Navbar/Menu.tsx
```diff
@@ -0,0 +1,40 @@
...
+    console.error("Navbar/Menu component error", error.message);
+  }
+
+  const menu = data?.menu?.items || [];
+
+  return (
+    <nav className={styles.nav}>
+      <ol>
+        {menu?.map((item) => {
+          if (!item) {
```
bmigirl, line 26: Same as above

components/Navbar/NavIconButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+import styles from "./Navbar.module.css"; // in the final version it should be imported from ui-kit package
+import Spyglass from "./spyglassIcon.svg"; // in the final version it should be imported from ui-kit package
+import User from "./userIcon.svg"; // in the final version it should be imported from ui-kit package
+
+interface NavIconButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
+  icon: "user" | "bag" | "spyglass" | "menu" | "close";
+  counter?: number;
+}
+
+const switchIcon = (icon: NavIconButtonProps["icon"]) => {
```
bmigirl, line 15: I think we could name this function a bit better. One thing is that whatever place uses this function doesn't need to be aware of how it's implemented - I'd rather call it `getIcon` and I think the name `icon` as a prop where it's actually `iconName` is a bit of a misleading name, since by the definition here (considering the name of this function) you kind of return an `icon` as an icon component, but also (according to current naming of the prop) pass an `icon` as a prop - this time as an icon name.

components/Navbar/UserMenu.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+function UserMenu({ className, ...rest }: UserMenuProps) {
+  const paths = usePaths();
+  const t = useIntl();
+  const { logout } = useAuth();
+  const { resetCheckoutToken } = useCheckout();
+  const router = useRouter();
+  const client = useApolloClient();
+
+  const onLogout = async () => {
```
bmigirl, line 26: Same as above

bmigirl (changes requested):

Einarsson:
@adrian-potepa Is this an "official" saleor storefront figma?

components/BurgerMenu/BurgerMenu.tsx
jwm0: After recent schema changes it should now be safe to omit

components/BurgerMenu/CollapseMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+            })}
+            onClick={() => setOpen(!open)}
+          >
+            {translate(data, "name")}
+            <ChebronDown />
+          </button>
+          {open && (
+            <div>
+              {data?.children?.map((sub) => (
+                <li key={sub?.id} className={styles["menu-link"]}>
```
jwm0, line 52: are there cases where `sub` is undefined - perhaps we can omit the optional chaining here

components/BurgerMenu/CollapseMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+                  {sub?.url ? (
+                    <a href={sub.url} target="_blank" rel="noreferrer">
+                      {sub?.name}
+                    </a>
+                  ) : (
+                    <Link href={getLinkPath(sub!, currentChannel.slug, currentLocale)} passHref>
```
jwm0, line 58: here we are making an assertion that sub is defined, see what the type should actually look like and type this code block properly

components/BurgerMenu/BurgerMenu.tsx
krzysztofwolski: Bless those changes

bmigirl (approved):

lib/auth.ts
```diff
@@ -0,0 +1,26 @@
...
+  const { logout } = useAuth();
+  const { resetCheckoutToken } = useCheckout();
+  const router = useRouter();
+  const client = useApolloClient();
+  const paths = usePaths();
+
+  const onLogout = async () => {
+    await logout();
+    await resetCheckoutToken();
+    await client.resetStore();
```
jwm0, line 19: some of these are not async, and I think we can push router immiedietly

jwm0 (approved):

szczecha (approved):

PR #198, opened 2022-05-04
Fix build and hydration

krzysztofwolski:
- drop build time pages generation and rely on ISR: pages are generated in the "blocking" mode so only the first render for the customers will be slower. Next ones will be taken from the cache.
- remove broken lighthouse action: we'll migrate to an external tool. The previous solution depended on an unmaintained community solution
- add nextjs build caching for CI: for the faster builds
- fix an issue with the hydration error on the checkout view: full-screen hydration error no longer displayed when loading the checkout view on the local development
- fix uncontrolled input error on the cart page: no console errors on the checkout view anymore
- font size fixes: bigger labels on the add to cart and checkout buttons
- hide the 'edit' button at the checkout view if the email is not set yet: since we do not use example email anymore by default, before the fix, edit buttons for address were visible even without email.


jwm0 (commented):
I'm unsure whether relying fully on ISR is the right call here, while the build might take longer it seems like it's a worthwhile investment _for some of the pages_ like categories and collections

components/checkout/BillingAddressSection.tsx
```diff
@@ -48,25 +48,28 @@ export function BillingAddressSection({ active, checkout }: BillingAddressSectio
...
-          {!!checkout.billingAddress && <AddressDisplay address={checkout.billingAddress} />}
-          <Button onClick={() => setEditing(true)}>{t.formatMessage(messages.changeButton)}</Button>
-        </section>
-      )}
+          </>
+        ) : (
+          <section className="flex justify-between items-center mb-4">
+            {!!checkout.billingAddress && <AddressDisplay address={checkout.billingAddress} />}
+            <Button onClick={() => setEditing(true)}>
+              {t.formatMessage(messages.changeButton)}
```
bmigirl, line 69: Just out of curiosity, why not do `const { formatMessage } = useIntl();` instead of creating a single letter variable name that doesn't matter here anyway?

bmigirl (approved):

krzysztofwolski:
@jwm0 discussed this with Timur earlier. The consensus was that approach to ISR is strictly related to the particular storefront implementation and the rate of data changes. For the demo purposes, dependent on it is ok and documentation should handle instructions on how to approach it for particular use cases.

components/checkout/BillingAddressSection.tsx
krzysztofwolski: Good idea, would like to make such change in the separate PR 👍

jwm0 (approved):

szczecha (approved):

Issue #203, opened 2022-05-08
Can't add items without images to the cart


bandieramonte:
### Summary

When browsing the product details for a certain item which has no media images attached to it, and clicking on the "Add to cart" button, I would expect the item to be added to the cart and to be able to see the checkout screen with such item added to the cart.

What actually happens upon clicking such button, is that the site crashes showing the following error:
_Unhandled Runtime Error
Error: Image is missing required "src" property. Make sure you pass "src" in props to the `next/image` component. Received: {}_

### Steps to reproduce

1. Add a product in the Dashboard without attaching any media images to it.
2. Go to the product details page for such product.
3. Click the "Add to cart" button.
4. The site crashes

### What OS are you using?

Windows

### Your `next.config.js` file

<details>
  <summary>`next.config.js`</summary>

```toml
const withBundleAnalyzer = require("@next/bundle-analyzer")({
  enabled: process.env.ANALYZE === "true",
});

const apiURL = new URL(process.env.NEXT_PUBLIC_API_URI);
const allowedImageDomains = process.env.NEXT_PUBLIC_ALLOWED_IMAGE_DOMAINS
  ? process.env.NEXT_PUBLIC_ALLOWED_IMAGE_DOMAINS.split(",")
  : [];
const imageConversionFormats = process.env.NEXT_PUBLIC_IMAGE_CONVERSION_FORMATS
  ? process.env.NEXT_PUBLIC_IMAGE_CONVERSION_FORMATS.split(",")
  : [];

module.exports = withBundleAnalyzer({
  reactStrictMode: true,
  swcMinify: true,
  images: {
    domains: [apiURL.hostname, ...allowedImageDomains],
    formats: imageConversionFormats,
  },
  async headers() {
    return [
      {
        source: "/(.*)",
        headers: [
          {
            key: "x-content-type-options",
            value: "nosniff",
          },
          { key: "x-xss-protection", value: "1" },
          { key: "x-frame-options", value: "DENY" },
          {
            key: "strict-transport-security",
            value: "max-age=31536000; includeSubDomains",
          },
        ],
      },
    ];
  },
  experimental: {
    reactRoot: true,
  },
});

```

</details>

### Builds logs (or link to your logs)

<details>
  <summary>Build logs</summary>

```
```

</details>

### Function logs

<details>
  <summary>Function logs</summary>

```
[RuntimeError.log](https://github.com/saleor/react-storefront/files/8646837/RuntimeError.log)

```

</details>

PR #204, opened 2022-05-08
Handle products without the images at cart and checkout 

bandieramonte:
This PR aims to fix issue [203](https://github.com/saleor/react-storefront/issues/203), so that items without media images can now be added to the cart.


matteius:
Not sure why the difference, but I am able to add products with out images to the cart.  For example:  https://mattscoinage.com/default-channel/en-US/products/test-stripe

bandieramonte:
> Not sure why the difference, but I am able to add products with out images to the cart. For example: https://mattscoinage.com/default-channel/en-US/products/test-stripe

Interesting. In any case, I think this new code adds a safety layer for cases without images - it doesn't hurt.

krzysztofwolski (approved):
LGTM 👍

krzysztofwolski:
Tested locally, and indeed it solves the issue with the products without the images :) 
Thanks for your contribution!

PR #212, opened 2022-05-11
PLP filters and redesign

krzysztofwolski:


components/ProductCollection/ProductCollection.tsx
```diff
@@ -11,20 +18,47 @@ import { messages } from "../translations";
...
+    first: perPage,
+    ...query,
+  };
+
+  if (sortBy?.field && sortBy?.direction) {
+    variables.sortBy = {
+      direction: sortBy.direction,
+      field: sortBy.field,
+    };
+  }
```
jwm0, line 50: This if can be omitted and the logic should be moved to the line above:
```
  const variables: ProductCollectionQueryVariables = {
    filter,
    first: perPage,
    ...query,
    ...(sortBy?.field && { field: sortBy.field })
    ...(sortBy?.direction && { direction: sortBy.direction }) 
  };
```

components/productList/FilteredProductList/SortingDropdown.tsx
```diff
@@ -0,0 +1,84 @@
...
+  for (const option of options) {
+    if (option.field === chosen?.field && option.direction === chosen?.direction) {
+      option.chosen = true;
+      isChosenSet = true;
+      break;
+    }
+  }
+  if (!isChosenSet) {
+    options[0].chosen = true;
+  }
```
jwm0, line 39: The logic is fine but I'd move it outside the component and call it: `getSortingOptions(chosen)`

components/productList/FilteredProductList/attributes.ts
```diff
@@ -0,0 +1,73 @@
...
+        label: `${attrName}: ${choiceName}`,
+        choiceSlug: value,
+        attributeSlug: filter.slug,
+      });
+    }
+  }
+  return pills;
+};
+
+export const parseQueryAttributeFilters = (query: string): UrlFilter[] => {
```
jwm0, line 37: There's this package here that simplifies entire query parsing and serialization: https://www.npmjs.com/package/qs

jwm0 (approved):

components/productList/FilteredProductList/attributes.ts
krzysztofwolski: Don't think it's worth adding to the bundle size (https://bundlephobia.com/package/qs@6.10.3). This solution is inspired by what has been used in the old storefront.

BTW pkg we are using for URL state has some serialization options, but URL will not be as easy to read as with this custom method.

jwm0: Fair enough, that's a pretty bloated package 👀

patrys:
Task linked: [SALEOR-7088 Release PLP](https://app.clickup.com/t/2549495/SALEOR-7088)

components/productList/FilteredProductList/sorting.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export interface SortingOption {
+  label: string;
+  field?: ProductOrderField;
+  direction?: OrderDirection;
+  chosen: boolean;
+}
+
+export const getSortingOptions = (chosen: UrlSorting | null) => {
```
bmigirl, line 15: Use map instead of lets + for. Also, since `chosen` is later on a boolean prop in the options object, the name `chosen` for what is in fact a urlSorting object is misleading. Let's call it `urlSorting` maybe?

pages/[channel]/[locale]/category/[slug].tsx
```diff
@@ -25,28 +33,70 @@ export const getStaticProps = async (context: GetStaticPropsContext) => {
...
+  attributeFiltersData,
+}: InferGetStaticPropsType<typeof getStaticProps>) {
+  const paths = usePaths();
+  const router = useRouter();
+
   if (!category) {
     return <Custom404 />;
   }
+
+  const subcategories = category.children?.edges.map((edge) => edge.node) || [];
```
Comment (bmigirl) [inline pages/[channel]/[locale]/category/[slug].tsx:73]: Why not add a util for extracting those from edges? It's repeated a few times here and there

components/productList/FilteredProductList/FilteredProductList.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  const removeAttributeFilter = (attributeSlug: string, choiceSlug: string) => {
+    const existingFilter = queryFilters.find((f) => f.slug === attributeSlug);
+    if (!existingFilter) {
+      return;
+    }
+
+    // if it is last choice value, remove whole attribute from the list
+    if (existingFilter.values.length === 1) {
+      const updatedFilters = queryFilters.filter((f) => f.slug !== attributeSlug);
```
bmigirl, line 88: `f` could be just named `filter`

components/productList/FilteredProductList/FilteredProductList.tsx
```diff
@@ -0,0 +1,216 @@
...
+      setQueryFilters(updatedFilters.length ? updatedFilters : null, {
+        scroll: false,
+        shallow: true,
+      });
+      return;
+    }
+
+    // if there are other values, just remove selected one
+    existingFilter.values = existingFilter.values.filter((v) => v !== choiceSlug);
```
bmigirl, line 97: Same as above

components/productList/FilteredProductList/FilteredProductList.tsx
```diff
@@ -0,0 +1,216 @@
...
+    // if given attribute and value has been already chosen, remove it
+    if (
+      pills.find((pill) => pill.attributeSlug === attributeSlug && pill.choiceSlug === choiceSlug)
+    ) {
+      removeAttributeFilter(attributeSlug, choiceSlug);
+      return;
+    }
+
+    // if attribute was not used before, add it
+    const existingFilter = queryFilters.find((f) => f.slug === attributeSlug);
```
bmigirl, line 114: Same as above

components/productList/FilteredProductList/FilteredProductList.tsx
```diff
@@ -0,0 +1,216 @@
...
+      attributes: queryFilters.filter((f) => f.values?.length),
+      ...(categoryIDs?.length && { categories: categoryIDs }),
+      ...(collectionIDs?.length && { collections: collectionIDs }),
+      ...(inStockFilter && { stockAvailability: "IN_STOCK" }),
+    });
+    // Eslint does not recognize stringified queryFilters, so we have to ignore it
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [inStockFilter, JSON.stringify(queryFilters), categoryIDs, collectionIDs]);
+
+  const removeAttributeFilter = (attributeSlug: string, choiceSlug: string) => {
```
bmigirl, line 80: All these helper functions could potentially either be in utils or moved to a hook (whatever is necessary) so this component doesn't keep all this mapping and filtering logic. Just a suggestion tho

pages/[channel]/[locale]/collection/[slug].tsx
```diff
@@ -25,29 +31,56 @@ export const getStaticProps = async (context: GetStaticPropsContext) => {
...
+    variables: {
+      ...contextToRegionQuery(context),
+      filter: {
+        inCollection: response.data.collection?.id,
+      },
+    },
+  });
+
+  let attributes: AttributeFilterFragment[] =
+    attributesResponse.data.attributes?.edges.map((e) => e.node) || [];
```
Comment (bmigirl) [inline pages/[channel]/[locale]/collection/[slug].tsx:49]: As above, a util to extract those would fit nicely here

bmigirl (changes requested):

components/productList/FilteredProductList/FilteredProductList.tsx
```diff
@@ -0,0 +1,218 @@
...
+    const newFilters = [];
+    for (const fa of queryFilters) {
```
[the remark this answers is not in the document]
krzysztofwolski: Approached it in a different way.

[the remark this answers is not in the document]
bmigirl: Reduce is designed for cases like this. What's the benefit of the approach you've chosen?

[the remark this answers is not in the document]
krzysztofwolski: The benefit was it was reimplemented before your review.
Finally, I've submitted your approach

bmigirl (approved):

szczecha (approved):

PR #217, opened 2022-05-13
Typography sizes

krzysztofwolski:
- add missing text-base classes
- modify rich text block

### DIFF
--- a/components/RichText/RichText.tsx
+++ b/components/RichText/RichText.tsx
@@ -1,32 +1,16 @@
 import Blocks from "editorjs-blocks-react-renderer";
 import React from "react";
 
+import { parseEditorJSData } from "@/lib/util";
+
 export interface RichTextProps {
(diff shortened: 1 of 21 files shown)


components/RichText/RichText.tsx
```diff
@@ -13,15 +13,24 @@ export function RichText({ jsonStringData }: RichTextProps) {
...
     return null;
   }
+
+  // Path for compatibility with data from older version od EditorJS
+  if (!data.time) {
+    data.time = Date.now().toString();
+  }
+  if (!data.version) {
+    data.version = "2.22.2";
+  }
```
jwm0, line 30: Move this to a local function outside the component `getParsedData` and call it in the component:
```
const data = getParsedData(jsonStringData);
```

jwm0 (changes requested):

jwm0 (approved):

PR #253, opened 2022-06-13
configuration of cypress, sample test to check if works

eCzerniak:
Adding cypress with sample test.


.gitignore
```diff
@@ -36,3 +36,7 @@ yarn-error.log*
 
 # localcache
```
michalina-graczyk, line 37: Do we need these comments?

.gitignore
```diff
@@ -36,3 +36,7 @@ yarn-error.log*
...
 .eslintcache
+
+#cypress
```
michalina-graczyk, line 40: same here ^

.gitignore
eCzerniak: this concept has already been addressed, so this question should be addressed to @krzysztofwolski or @zaiste

zaiste (changes requested):
@eCzerniak Great initiative! Would it be possible to default to TypeScript when writing those tests? This could simplify the management of this code base

michalina-graczyk (approved):

zaiste (approved):

PR #261, opened 2022-06-28
Add tests for filters srs 03

eCzerniak:
Adding tests for filters. 
Adding `data-testid` to elements.

### DIFF
--- a/cypress/e2e/filteringProducts.cy.ts
+++ b/cypress/e2e/filteringProducts.cy.ts
@@ -10,15 +10,15 @@ describe("Filter products", () => {
 
   it("should filter products by variant attribute SRS_0306", () => {
     waitForProgressBarToNotBeVisible();
-    filterProducts(NAVIGATION.categoriesList, CATEGORY.categoryTitle);
-    cy.get(FILTERS.filtersMenu).first().click();
+    filterProducts(NAVIGATION.categoriesListButtons, CATEGORY.categoryTitle);
+    cy.get(FILTERS.filtersMenuButtons).first().click();
     filterProducts(FILTERS.filterList, FILTERS.filterPill);
   });
 
   it("should clear selected filters SRS_0308", () => {
     waitForProgressBarToNotBeVisible();
-    filterProducts(NAVIGATION.categoriesList, CATEGORY.categoryTitle);
-    cy.get(FILTERS.filtersMenu).first().click();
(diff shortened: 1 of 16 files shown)


patrys:
Task linked: [SALEOR-7465 SRS_03 - Add tests for filters in SRS](https://app.clickup.com/t/2549495/SALEOR-7465)

cypress/elements/navigation.ts
```diff
@@ -0,0 +1,3 @@
+export const NAVIGATION = {
+  categoriesList: "[data-testid*='categoriesList']",
```
karola312, line 2: Wouldn't be better to name it the categoryListButtons? And the same with other elements?

cypress/support/shared.ts
```diff
@@ -0,0 +1,14 @@
+import { SHARED } from "cypress/elements/shared";
+
+export function waitForProgressBarToNotBeVisible() {
+  cy.get(SHARED.spinner).should("not.exist");
+}
+
+export function filterProducts(firstElement, secondElement) {
```
karola312, line 7: Variables firstElement, secondElement should be better named. Maybe filterProductsBy, and selectedFilter

cypress/e2e/filteringProducts.cy.ts
```diff
@@ -0,0 +1,30 @@
...
+describe("Filter products", () => {
+  beforeEach(() => {
+    cy.visit("/");
+  });
+
+  it("should filter products by variant attribute SRS_0306", () => {
+    waitForProgressBarToNotBeVisible();
+    filterProducts(NAVIGATION.categoriesList, CATEGORY.categoryTitle);
+    cy.get(FILTERS.filtersMenu).first().click();
+    filterProducts(FILTERS.filterList, FILTERS.filterPill);
```
karola312, line 15: Don't you want to check here if the graphql request for filtered products was sent and if all products from the response are displayed?

eCzerniak: I was thinking about that and it was little bit hard for me in this moment and we can do this if you would like to help me with that ;).

cypress/support/shared.ts
michalina-graczyk: yup, we should avoid using names which doesn't describe what the first element is

szczecha (approved):

michalina-graczyk (approved):

PR #262, opened 2022-07-05
Add test for add variant to cart

szczecha:
- adding a test of adding a variant to the cart
- adding `data-testid` to elements
- standardization of the names of exports
- adding `waitForSearchedProducts` to the search test

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -12,5 +12,8 @@ export default defineConfig({
   e2e: {
     baseUrl: "http://localhost:3001/",
     specPattern: "cypress/e2e/**/*.cy.{js,jsx,ts,tsx}",
+    env: {
+      API_URI: "https://vercel.saleor.cloud/graphql/",
+    },
   },
 });
--- a/cypress/e2e/addVariantToCart.cy.ts
+++ b/cypress/e2e/addVariantToCart.cy.ts
@@ -1,20 +1,19 @@
 import { CART } from "cypress/elements/cart";
-
-import { PRODUCT } from "../elements/product";
-import { SHARED } from "../elements/shared";
-import { productsToSearch } from "../fixtures/search";
-import { navigateAndSearch, waitForSearchedProducts } from "../support/pages/search";
+import { PRODUCT } from "cypress/elements/product";
+import { SHARED } from "cypress/elements/shared";
+import { productsToSearch } from "cypress/fixtures/search";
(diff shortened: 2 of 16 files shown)


patrys:
Task linked: [SALEOR-7521 Add test for SRS_0202 should select a variant and add a product to the chart](https://app.clickup.com/t/2549495/SALEOR-7521)

cypress/support/pages/search.ts
```diff
@@ -1,10 +1,22 @@
...
+  cy.intercept("POST", Cypress.env("API_URI"), (req) => {
+    if (
+      req.body.operationName === "ProductCollection" &&
+      req.body.variables.filter.search === typedText
+    )
+      req.alias = "SearchResult";
+    req.continue();
+  });
+  cy.wait("@SearchResult");
+}
```
karola312, line 22: This can fail. First, we should intercept the request, then we should make an action that will trigger that request, and after that wait for an aliased request. In this case, everything is passing because it takes w while, when searching for products is sending the request

cypress/e2e/addVariantToCart.cy.ts
```diff
@@ -0,0 +1,52 @@
...
+            addItemToCart(PRODUCT.variant, selectedVariantName);
+          })
+          .get(CART.cartProduct)
+          .first()
+          .should("contain.text", selectedProductName);
+      });
+  });
+
+  it("should add product without variants to the cart SRS_0203", () => {
+    const product = productsToSearch.productWithoutVariants;
```
eCzerniak, line 40: Add an empty line.
```suggestion
    const product = productsToSearch.productWithoutVariants;
    
```

cypress.config.ts
```diff
@@ -12,5 +12,8 @@ export default defineConfig({
   e2e: {
     baseUrl: "http://localhost:3001/",
     specPattern: "cypress/e2e/**/*.cy.{js,jsx,ts,tsx}",
+    env: {
+      API_URI: "https://vercel.saleor.cloud/graphql/",
```
eCzerniak, line 16: Shouldn't we take this env from `.env` file?

eCzerniak (changes requested):

mmiszy:
Sorry, this PR was closed automatically due to us reorganising the repo. Would you mind reapplying your changes? Thanks for the contribution!

PR #265, opened 2022-07-08
cleanup in tests

eCzerniak:
- adding `' '` characters at `data-testid` selectors,
- adding empty lines between variables, 
- changing names of the files `shared` -> `shared-elements` and `shared` -> `shared-operations`
- adding scripts `cy:run` and `cy:open`

### DIFF
--- a/cypress/e2e/search.cy.ts
+++ b/cypress/e2e/search.cy.ts
@@ -5,16 +5,15 @@ import { productsToSearch } from "../fixtures/search";
 import { navigateAndSearch } from "../support/pages/search";
 
 describe("Search for products", () => {
-  let typedText;
-
   beforeEach(() => {
     cy.visit("/");
   });
 
   it("should search for products SRS_0405", () => {
-    typedText = productsToSearch.polo;
-    navigateAndSearch(typedText);
-    cy.url().should("include", `/search?q=${typedText}`);
+    const searchQuery = productsToSearch.polo;
+
+    navigateAndSearch(searchQuery);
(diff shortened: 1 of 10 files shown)


szczecha (approved):

patrys:
Task linked: [SALEOR-7582 Cleanup of SRS tests](https://app.clickup.com/t/2549495/SALEOR-7582)

cypress/e2e/search.cy.ts
```diff
@@ -1,15 +1,16 @@
...
 import { SEARCH_PAGE_SELECTORS } from "../elements/search-page";
-import { SHARED } from "../elements/shared";
+import { SHARED_ELEMENTS } from "../elements/shared-elements";
 import { productsToSearch } from "../fixtures/search";
 import { navigateAndSearch } from "../support/pages/search";
 
-let typedText;
-
 describe("Search for products", () => {
+  let typedText;
```
krzysztofwolski, line 8:
```suggestion
  const searchQuery;
```

krzysztofwolski (approved):

cypress/e2e/search.cy.ts
eCzerniak: @szczecha  What do you think about that?

szczecha: IMO it should stay as 'let' because depending on the test case I use different data for this variable. 
About renaming to `searchQuery` actually fits better here, but we will need to change it also in the test body.

karola312 (approved):

PR #268, opened 2022-07-12
Fix Double 404 Error Navbar

RyanTamulevicz:
Simply just removing the Navbar from showing up twice. An example can be seen here https://demo.saleor.io/default-channel/en-US/collection/top-artistss


mmiszy:
Sorry, this PR was closed automatically due to us reorganising the repo. Would you mind reapplying your changes? Thanks for the contribution!

PR #276, opened 2022-07-28
Lock pnpm version in Dockerfile to 7.5.2

lkostrowski:
In the future this should prevent unexpected behavior, when image is downloading newer version than we tested with

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -2,7 +2,7 @@ FROM node:16
 WORKDIR /app
 
 # Setup pnpm package manager
-RUN npm install -g pnpm@7.6.0
+RUN npm install -g pnpm@7.5.2
 
 # Setup proxy to API used in saleor-platform
 RUN apt-get update && apt-get install -y nginx
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -2,7 +2,7 @@ FROM node:16
 WORKDIR /app
 
 # Setup pnpm package manager
-RUN npm install -g pnpm
+RUN npm install -g pnpm@7.6.0
 
 # Setup proxy to API used in saleor-platform
 RUN apt-get update && apt-get install -y nginx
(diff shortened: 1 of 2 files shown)


Dockerfile.dev
```diff
@@ -2,7 +2,7 @@ FROM node:16
 WORKDIR /app
 
 # Setup pnpm package manager
-RUN npm install -g pnpm
+RUN npm install -g pnpm@7.6.0
```
mmiszy, line 5: I think Vercel uses 7.5.2 atm so we might want to stick to that for consistency?
Last week we had a problem with deployments because we accidentally updated pnpm to 7.6.0 locally but Vercel is lagging behind.

```suggestion
RUN npm install -g pnpm@7.5.2
```

lkostrowski: Yes, it can help us see problems earlier (broken on Vercel -> we should check Docker too)

krzysztofwolski:
@lkostrowski maybe we should add ignore flag instead of forcing the exact version?

lkostrowski:
> @lkostrowski maybe we should add ignore flag instead of forcing the exact version? https://pnpm.io/errors#err_pnpm_peer_dep_issues

I don't think this is good idea. 

There are reasons why package managers are showing errors - we should maintain dependencies versions to work correctly.
If we ignore errors, we can face some silent bugs

The exact version will prevent case from broken Storefront, caused by Storefront using v6, Dockerfile was installing V7.
I'd like to prevent this in the future. Without locked version, PNPM can install `v8` and break it again.

zaiste:
The suggested way (their docs) to install `pnpm` is by using [the standalone script](https://pnpm.io/installation#using-a-standalone-script). This also locks the version, but it also provides a pre-built binary. A nitpick, but it may help to make it more stable.

mmiszy:
@zaiste they suggest 5 different ways of installing `pnpm`. What are the benefits of using a script instead of npm here?
BTW, we're using [Corepack](https://pnpm.io/installation#using-corepack) on Vercel.

Issue #307, opened 2022-07-26
TEST_* env variables are loaded in development


witoszekdev:
## Steps to reproduce the problem
1. Remove `TEST_*` env variables from `.env.local` (if they exist)
2. Run `pnpm dev --filter=saleor-app-checkout`
3. Make a request to any API route

## Current result
We get an error:

```
Invariant Violation: TEST_MOLLIE_KEY is not defined
```

## Expected behavior
`TEST_*` variables are not loaded in production or development build, only in tests.

## Environment

Not relevant

## Additional context
Reported by @bmigirl: https://github.com/saleor/saleor-checkout/pull/174#issuecomment-1195547660

Issue #310, opened 2022-07-27
Remove MSW and fix Polly recordings


witoszekdev:
# Requests made by payment provider SDKs are not recorded by Polly.js
### Case 1
1. Run test with `POLLY_MODE=record`, e.g. `POLLY_MODE=record pnpm run test`

### Case 2
1. Change values for `TEST_MOLLIE_KEY` to be invalid Mollie API key
2. Run tests in replay mode: `pnpm run test`

## Current result

Requests made by Mollie SDK are not recorded by Polly.js

In **case 1**: Requests made to Mollie SDK won't be saved in `/recordings`

In **case 2**: Requests will be made to the real Mollie SDK and fail because the API key is invalid.

## Expected behavior

All requests are intercepted by Polly.js and recorded.

#### System:
 - OS: macOS 12.4
 - CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
 - Memory: 118.25 MB / 16.00 GB
 - Shell: 5.8.1 - /bin/zsh
#### Binaries:
 - Node: 18.2.0 - /usr/local/bin/node
 - Yarn: 1.22.18 - /usr/local/bin/yarn
 - npm: 8.9.0 - /usr/local/bin/npm
 - Watchman: 2022.05.16.00 - /usr/local/bin/watchman
#### Browsers:
 - Brave Browser: 103.1.41.100
 - Chrome: 103.0.5060.134
 - Firefox Developer Edition: 101.0
 - Safari: 15.5

#### pnpm
- pnpm: 7.5.2

## Additional context

When I've changed Polly.js adapter in `test-utils.ts` to `@pollyjs/adapter-node-http` and harcoded responses from requests made to Saleor API (private app metadata) in `backend/payments/providers/mollie/utils.ts:15` the requests to Mollie API were intercepted properly by Polly.js.

However this breaks manual mocks for Saleor API - they are no longer handled and fail.

We might solve this by getting rid of msw and instead mocking functions that return make requests to retreive app's metadata from Saleor.

mmiszy:
We need to get rid of MSW (Mock Service Worker) and revamp our tests setup.

PR #318, opened 2022-07-29
Fix invariant errors in API routes for TEST_* vars

witoszekdev:
Reopened PR from https://github.com/saleor/saleor-checkout/pull/189

This PR fixes saleor/react-storefront#307 issue with invariant errors caused by missing TEST_* env vars - a module that imported fixtures was used in non-testing code which caused this error

### DIFF
--- a/apps/saleor-app-checkout/.env
+++ b/apps/saleor-app-checkout/.env
@@ -1,3 +1,14 @@
 NEXT_PUBLIC_SALEOR_API_URL=$SALEOR_API_URL
 NEXT_PUBLIC_CHECKOUT_APP_URL=$CHECKOUT_APP_URL
-SALEOR_APP_TOKEN=
+
+# Default values for tests
+# They will work only when replaying requests with Polly.js
+# To record new requests, use real credentials in .env.local
+TEST_MOLLIE_KEY=mollie_test_key
+TEST_MOLLIE_PROFILE_ID=mollie_profile_id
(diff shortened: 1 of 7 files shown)


apps/saleor-app-checkout/.env
```diff
@@ -1,3 +1,14 @@
 NEXT_PUBLIC_SALEOR_API_URL=$SALEOR_API_URL
 NEXT_PUBLIC_CHECKOUT_APP_URL=$CHECKOUT_APP_URL
-SALEOR_APP_TOKEN=
+
+# Default values for tests
+# They will work only when replaying requests with Polly.js
+# To record new requests, use real credentials in .env.local
+TEST_MOLLIE_KEY=mollie_test_key
```
mmiszy, line 7: Are you sure this will work? I think I came across a problem where Mollie does some kind of key validation in their SDK and they said it was malformed.

witoszekdev: I've checked it and it's not a problem with Mollie SDK - they don't do any validation in the SDK itself. It's actually a problem with our request interception: https://github.com/saleor/react-storefront/issues/310. Right now Polly.js doesn't intercept requests properly and the tests might fail if the `TEST_MOLLIE_KEY` is not configured with a real Mollie API key. I made a draft PR to fix this: https://github.com/saleor/react-storefront/pull/325

The goal of this PR hoewever was to fix the production build which was caused by importing `testingVars` so I guess we can leave it for now 😅

mmiszy (approved):

PR #321, opened 2022-07-29
Handle refunds from Adyen

witoszekdev:
This PR improves refund handling with Adyen
- Added more tests
- Handling of different edge cases (ex. `REFUND_FAILED` event)
- Handling of chargebacks (they're treated as refunds in Saleor)
- Rewrote Adyen webhook handler to use middlewares and saleor-app-sdk

TODO:
- [x] Add assertions for Saleor API calls in `__tests__/pages/api/webhooks/adyen.test.ts`
- [x] Change fixture for Adyen domain name: https://github.com/saleor/react-storefront/blob/f335b48882bedfcb3ec5e7f9375d3e8710a65577/apps/saleor-app-checkout/__tests__/backend/payments/providers/adyen/middlewares.test.ts#L24


apps/saleor-app-checkout/__tests__/backend/payments/providers/adyen/utils.test.ts
```diff
@@ -2,53 +2,379 @@ import {
...
+  it("throws an error if amount is missing", () => {
+    const notification = {
+      ...prepareAdyenNotification(12_00, EventCodeEnum.Authorisation),
+      amount: undefined,
+    };
+
+    expect(() => {
+      return getTransactionAmountFromAdyen(
+        // @ts-expect-error
+        notification,
```
mmiszy, line 113: Is it possible to assert the type here? As in:
```suggestion
        notification as Types.notification.NotificationRequestItem,
```

This way if we ever modify the implementation of `getTransactionAmountFromAdyen` there's higher chance we'll spot the error.

apps/saleor-app-checkout/__tests__/pages/api/webhooks/adyen.test.ts
```diff
@@ -0,0 +1,261 @@
+/* eslint-disable require-await */
```
mmiszy, line 1: Why tho

mmiszy: If you think it makes sense, we can disable it for all tests files via `overrides` in eslintconfig

apps/saleor-app-checkout/test-utils.ts
```diff
@@ -11,6 +11,7 @@ import { MockedRequest } from "msw";
 declare module "next" {
   type NextApiResponse = {
     _getJSONData: <T extends Object>() => T;
+    _getData: <T extends string>() => T;
   } & NextApiResponse;
 }
```
mmiszy, line 16: We shouldn't be doing this. It affects the whole codebase, not just tests. Let's remove this type and instead assert:

```suggestion
export type TestNextApiResponse = NextApiResponse & {
  _getJSONData: <T extends Object>() => T;
  _getData: <T extends string>() => T;
};
```

apps/saleor-app-checkout/backend/environment.ts
```diff
@@ -7,6 +7,12 @@ const maskToken = (token: string) => "*".repeat(Math.max(token.length - 4, 0)) +
 
 export const getAuthToken = () => {
   let token;
+
+  if (IS_TEST) {
+    // Allows to use real appToken to record requests in development
+    token = "TEST";
+  }
```
mmiszy, line 14: Why not just use `serverEnvVars.appToken` in tests? Is this condition even necessary?

apps/saleor-app-checkout/backend/payments/providers/adyen/middlewares.ts
```diff
@@ -0,0 +1,114 @@
...
+        "Cannot fetch Adyen API configuration"
+      );
+    }
+
+    const {
+      paymentProviders: { adyen },
+    } = settings;
+
+    const keys = new Set(Object.keys(adyen));
+    const expectedKeys: Array<keyof AdyenRequestContext> = [
```
mmiszy, line 34: If we ever add anything to `AdyenProviderSettingID` this code should fail – but it won't.

We should modify how `AdyenProviderSettingID` is created and instead infer it from a tuple like so:

```ts
export const adyenProviderSettingIDs = [
  "hmac",
  "apiKey",
  "username",
  "password",
  "clientKey",
  "merchantAccount",
] as const;
export type AdyenProviderSettingID = typeof adyenProviderSettingIDs[number];
```

apps/saleor-app-checkout/backend/payments/providers/adyen/utils.ts
```diff
@@ -1,32 +1,43 @@
...
+    if (operation === OperationsEnum.Refund) {
       return "REFUND";
     }
 
-    if (operation === Types.notification.NotificationRequestItem.OperationsEnum.Cancel) {
+    if (operation === OperationsEnum.Cancel) {
       return "VOID";
     }
 
     throw "OperationsEnum out of bounds";
```
mmiszy, line 38: Since you're already refactoring this, we could us a `switch` or an object to make it more concise. Like:

```ts
switch (operation) {
  case OperationsEnum.Capture:
    return "CHARGE";
  case OperationsEnum.Refund:
    return "REFUND";
  case OperationsEnum.Cancel:
    return "VOID";
  default:
    throw new Error("OperationsEnum out of bounds");
}
```

apps/saleor-app-checkout/backend/payments/providers/adyen/utils.ts
```diff
@@ -44,12 +55,209 @@ export const getLineItems = (lines: OrderFragment["lines"]): Types.checkout.Line
...
+
+  return "SUCCESS";
+};
+
+export type TransactionAmounts = Pick<
+  TransactionUpdateInput,
+  "amountRefunded" | "amountAuthorized" | "amountCharged" | "amountVoided"
+>;
+
+const nonNegative = (amount: number) => (amount < 0 ? 0 : amount);
```
mmiszy, line 86: Probably doesn't matter but this function fails to do its job for `-0` 🙃  Surprisingly, the below change fixes the problem:
```suggestion
const nonNegative = (amount: number) => (amount <= 0 ? 0 : amount);
```

apps/saleor-app-checkout/backend/payments/providers/adyen/utils.ts
```diff
@@ -44,12 +55,209 @@ export const getLineItems = (lines: OrderFragment["lines"]): Types.checkout.Line
...
+    transaction?.voidedAmount?.currency,
+    transaction?.chargedAmount?.currency,
+    transaction?.authorizedAmount?.currency,
+  ];
+
+  // find first existing currency
+  return currencies.find((item) => typeof item === "string");
+};
+
+export const getTransactionAmountFromAdyen = (
```
mmiszy, line 104: Looks like this function could be split into a few smaller ones. Right now it's hard for me to follow what's happening.

apps/saleor-app-checkout/backend/payments/providers/adyen/utils.ts
```diff
@@ -44,12 +55,209 @@ export const getLineItems = (lines: OrderFragment["lines"]): Types.checkout.Line
...
+    typeof notification.amount.currency !== "string" ||
+    typeof notification.amount.value !== "number"
+  ) {
+    console.error("(Adyen webhook) Notification without amount or currency");
+    throw new Error("Notification doesn't contain amount or currency");
+  }
 
-  if (chargedAmount < 0) {
-    throw "Amount after refund cannot be negative";
+  if (transactionCurrency && transactionCurrency !== notification.amount.currency) {
```
mmiszy, line 135: What if there's no `transactionCurrency` ? Shouldn't we fail?

apps/saleor-app-checkout/jest.config.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable import/no-anonymous-default-export */
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
```
mmiszy, line 2: Shouldn't this be on line 8?

apps/saleor-app-checkout/jest.config.js
```diff
@@ -10,14 +11,25 @@ module.exports = {
...
+    "ts-jest": {
+      useESM: true,
+      tsconfig: {
+        ...compilerOptions,
+        jsx: "react-jsx",
+      },
+    },
+  },
   moduleNameMapper: {
+    "^(\\.{1,2}/.*)\\.js$": "$1",
```
mmiszy, line 25: What are we doing here? 🤔

mmiszy (changes requested):

apps/saleor-app-checkout/backend/payments/providers/adyen/utils.ts
witoszekdev: ah yes, the quirky `-0` 😅  - thanks good catch

apps/saleor-app-checkout/jest.config.js
witoszekdev: yes, it should - my bad, fixed

apps/saleor-app-checkout/__tests__/pages/api/webhooks/adyen.test.ts
witoszekdev: Nope, it doesn't make sense, we shouldn't disable this rule - I've replaced the `mockImplementationOnce` with `mockResolvedValueOnce`

apps/saleor-app-checkout/__tests__/backend/payments/providers/adyen/utils.test.ts
witoszekdev: no, we can't assert *in this place* - because we check for runtime errors TypeScript will catch. I've added however an assertion on the fields we don't check for errors (so everything, except `amount`)

mmiszy: I don't think I understand what you mean by `runtime errors TypeScript will catch`. Can you elaborate?

witoszekdev: Ah, sorry - the previous answer doesn't really make sense:
- TypeScript will check statically errors in our code, in this line of test it will raise an issue because we're passing a notification object **without amount**.
- We do that on purpose, we want to check if the function handles a case when amount is missing **in runtime**

mmiszy: Gotcha! So the error is actually expected.

apps/saleor-app-checkout/backend/environment.ts
witoszekdev: You're right, I've moved the logic of setting env var into `setupTestsBeforeEnv` file. We do the same thing for encryption secret

apps/saleor-app-checkout/backend/payments/providers/adyen/utils.ts
witoszekdev: No, if a transaction doesn't have currency set, then this means:
- it doesn't have any amount set yet
- there's no transaction, and we need to create it

This should fail only if there **is** a currency, and it doesn't match what came from Adyen. It's unlikely to happen at all, but better safe than sorry 😅

apps/saleor-app-checkout/backend/payments/providers/adyen/middlewares.ts
witoszekdev: That's perfect idea 🙌🏻 
I totally forgot about this type, but it totally makes sense to use it here

apps/saleor-app-checkout/test-utils.ts
witoszekdev: Done, I've changed it

apps/saleor-app-checkout/jest.config.js
witoszekdev: Honestly, I'm not 100% sure 🙈. I think it was supposed to strap `.js` from file names - I've tried while I was debugging the issues with `ts-jest` and must have left that by accident. But since we use the preset from `ts-jest` now, it doesn't make sense to have this anymore.

witoszekdev: I've removed that

apps/saleor-app-checkout/backend/payments/providers/adyen/utils.ts
witoszekdev: I've tried to refactor this a big - especially the conditions to validate the notification and currency, but I don't really have a better idea how to improve the big chunk of `switch` statement. I guess we could refactor each condition into a separate function, but there wouldn't be any overlap in functionality (maybe with `CancelOrRefund` case) - each case needs to return different data.

packages/checkout-common/package.json
```diff
@@ -23,7 +23,9 @@
     "@rollup/plugin-commonjs": "^22.0.0",
     "@rollup/plugin-node-resolve": "^13.2.1",
     "@rollup/plugin-typescript": "^8.3.2",
+    "esbuild": "^0.15.0",
```
mmiszy, line 26: Why?

packages/checkout-common/index.d.ts
```diff
@@ -0,0 +1,82 @@
+import { SvgIconTypeMap } from "@material-ui/core";
```
mmiszy, line 1: This file shouldn't be here

packages/checkout-common/rollup.config.js
```diff
@@ -0,0 +1,33 @@
...
+        format: "cjs",
+        sourcemap: true,
+      },
+      {
+        file: packageJson.module,
+        format: "es",
+        sourcemap: true,
+      },
+    ],
+    plugins: [esbuild(), typescript({ tsconfig: "./tsconfig.json" })],
```
mmiszy, line 26: Why are we not using the same rollup config as other packages in this repo?

mmiszy (changes requested):

packages/checkout-common/index.d.ts
witoszekdev: whoops, my bad - removed

packages/checkout-common/rollup.config.js
witoszekdev: Honestly: I've used it when I firstly tried to debug why the build was failing, and I couldn't get it to work, the erorr messages were too cryptic for me 😅.

I've tried simpler config, just cjs and esbuild and it started to work for me. I've left it since I though that maybe we should use esbuild rather than multiple plugins? We already use it to build dashboard and macaw-ui. The performance difference is massive, although for this small package this doesn't even matter since we literaly export 1 variable and the rest is just types.

I've copied config from ui-kit `-` postcss

packages/checkout-common/package.json
witoszekdev: https://github.com/saleor/react-storefront/pull/321#discussion_r943238627

packages/checkout-common/package.json
```diff
@@ -6,6 +6,9 @@
   "module": "dist/esm/index.js",
   "types": "./dist/index.d.ts",
   "scripts": {
+    "build": "rollup -c",
```
mmiszy, line 9: Something is wrong because these scripts are already there on main: https://github.com/saleor/react-storefront/blob/382c0a3c484cf2c56a8952eeda0f53b56121d3ad/packages/checkout-common/package.json#L11

mmiszy (changes requested):
Can you also revert changes to the lockfile?

packages/checkout-common/package.json
witoszekdev: my bad, I didn't resolve the conflicts properly

mmiszy, line 9:
```suggestion
    "build": "NODE_ENV=production rollup -c",
```

We rely on this inside rollup.config.js

mmiszy (dismissed):

szczecha (approved):

mmiszy (approved):

PR #325, opened 2022-08-01
[On hold] Bugfix: record requests made to Mollie in Polly.js

witoszekdev:
Attempt to fix https://github.com/saleor/react-storefront/issues/310 - instead of using msw, we mock methods that return app metadata from Saleor and use `@pollyjs/adapter-node-http` for intercepting requests in Polly.js


mmiszy:
We should try using pollyjs#intercept method to replace MSW.

PR #335, opened 2022-08-02
Prepare Dockerfile exposing storefront for saleor-platform

magul:


Dockerfile.dev
```diff
@@ -6,14 +6,14 @@ RUN npm install -g pnpm@7.5.2
```
lkostrowski, line 6: @saleor/checkout & @magul 

How about we make 1-Dockerfile per package? 
We eventually need one for Checkout - do we want to have 2 services in 1 Dockerfile (doesn't seem right).

Maybe we could have 
`/Dockerfile.storefront` &
`/Dockerfile.checkout`

or 

`/apps/storefront/Dockerfile` &
`apps/checkout/Dockerfile`

It's on docker-compose to decide what services to include

If we decide now, we will avoid further braking changes in platform

What do you think?

Dockerfile.dev
```diff
@@ -6,14 +6,14 @@ RUN npm install -g pnpm@7.5.2
...
-ENV API_URI ${API_URI:-http://localhost:8000/graphql/}
+RUN pnpm install
+
+
+ARG SALEOR_API_URL
+ENV SALEOR_API_URL ${SALEOR_API_URL:-http://localhost:8000/graphql/}
 
 EXPOSE 3000
-CMD pnpm dev -- --hostname 0.0.0.0
+CMD pnpm turbo run dev --parallel --cache-dir=.turbo
```
mmiszy, line 19: I'm wondering, why do we use `dev` command instead of `build` + `start`?

witoszekdev: @magul this file is named `Dockerfile.dev` - dev which means is's for development? 🤔 Shouldn't we use something else for local development and Docker for production deployments?

In this configuration if you change something in the project's code you'd still have to re-build the docker image on every change. If that's our use case then maybe we should use [bind mounts](https://docs.docker.com/storage/bind-mounts/) instead of `COPY`

lkostrowski: Maybe we should have 2 images, one development with volums, second production with COPY ?

witoszekdev: yup, for production `COPY` makes more sense

magul: @mmiszy:
> I'm wondering, why do we use `dev` command instead of `build` + `start`

I believe that will lower an entry level for exploring storefront, when fetched for the fist time inside `saleor-platform`.
My understanding of a purpose of `saleor-platform` is that you can just fetch it, build it, start it and start playing immediately with a source code (which will be shared between docker container and host machine using volumes).
Having `dev` instead `build` and `start` allow it to be hot-reloaded (as I correctly understand, and assuming that socket responsible for that will be also available from outside of a container).

@witoszekdev @lkostrowski  answering all your comments:
 * I need to `COPY` these assets during image build phase, as I need to be able to install npm dependencies there; we can have a conversation whether we want that to be done in entrypoint and making startup of these apps slower there (just be aware that I'm not changing this behaviour - it worked that way in pre-mono epoch also)
 * bind volumes were/are used in that way in pre-mono setup and I do intend to use them the same way (see: https://github.com/saleor/saleor-platform/blob/main/docker-compose.yml#L46, it will require changes, but I want to deliver that after figuring out how to encapsulate new, mono- storefront in docker)
 * beyond `saleor-platform` I dont' recall other development-centered usage for that `Dockerfile`, we could have a conversation about production-grade docker image, but I  don't know if we accutaly encorage people to use it as it is in their projects and want to keep that out of scope of this pull request (as that's yet another major conversation, how we want to support our users in that regard). One thing - is it used in demo e.g.? @witoszekdev maybe somebody from the cloud team know, what do we use there and how it's built?

Dockerfile.dev
magul: I was thinking about delivering `Dockerfile`, that will actually run storefront only, but as it relies on multiple of other npm workspaces, ultimate I've just:
```
COPY . .
```
assuming that could be considered good first approximation of what we want to achieve here, but I'm open for suggestions, how to actually deliver so fine-grained docker images without extensive overhead to be maintained.

Dockerfile.dev
witoszekdev: @magul Storefront repo is deployed on Vercel, it's not using Dockerfile in production. The `demo.saleor.io/graphql` instance is managed by cloud team, I don't know whay they use for that 😅 - it's special because it masks PII and the dashboard is read only.

magul: >  Storefront repo is deployed on Vercel
So maybe we don't need to provide production-grade docker image for that?

> demo.saleor.io/graphql
there's GraphQL playground/endpoint, but I asking about storefront that consumes that at https://demo.saleor.io/default-channel/en-US ? Is that this `react-storefront`? If cloud team support it and there's SSR involved, so they should have a way to deploy that (as we use Kubernetes, I'm assuming there's a `Dockerfile` somewhere). Also - are they aware that monorepo change and deployment pipeline reflects that already?

lkostrowski: [Here](https://github.com/saleor/saleor-platform/discussions/166) is discussion thread we can have regarding deployment strategy. At the moment let's fix local dev setup

magul: > Here is discussion thread we can have regarding deployment strategy. At the moment let's fix local dev setup

That's even more complex topic, as (as far as I understand) storefront is something unique for a store, so don't know if it should be considered as fixed part of Saloer suite of things that you want to deploy.
But yes, let's park that conversation for now.

Dockerfile.dev
lkostrowski: I think it's OK to copy everything (can be optimized later). My idea is to start only single service. 

Also, copy & install can be cached on docker layer AFAIK so only mainly CMD part will differ, right?

Dockerfile.dev
magul: I've just read conversation about issue in CLI regarding deploying storefront, so I'm withdrawing that claim:
> That's even more complex topic, as (as far as I understand) storefront is something unique for a store, so don't know if it should be considered as fixed part of Saloer suite of things that you want to deploy.

Silly me!

magul:
@lkostrowski @mmiszy @witoszekdev I believe I addressed all your comments - should I deliver something else here? or can we just merge it?

lkostrowski (approved):
I don't like idea of single Dockerfile running few projects in single container. However, we need to merge it fast, because we have regression and saleor-platform is not working correctly.

So I approve, hopefully lets merge it and we will think about next increment

@mmiszy cc

mmiszy:
For some reason tests are failing as if the emv variables were not set. It must be unrelated to this PR so I'll merge it anyway.
(cc: @witoszekdev )

witoszekdev:
The env vars were not set, because the PR was made from a fork - it's a security measure in order to prevent leaks 😉

magul:
> The env vars were not set, because the PR was made from a fork - it's a security measure in order to prevent leaks 😉

Well spotted @witoszekdev ! I don't have push access to that repository 🤷🏿

PR #346, opened 2022-08-03
Add of french locale (#278)

mguldner:
Add of french locale.
Resolving dicsussion #278


bmigirl:
@timuric can we deploy this?

mmiszy (dismissed):

PR #347, opened 2022-08-03
Stripe [WIP]

mmiszy:


apps/checkout/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "extends": "tsconfig/nextjs.json",
   "compilerOptions": {
-    "target": "ES2022",
-    "lib": ["DOM", "es2022"],
+    "target": "ES2021",
```
mmiszy, line 4: Unifying this everywhere

apps/saleor-app-checkout/package.json
```diff
@@ -92,6 +93,6 @@
     "setup-polly-jest": "^0.11.0",
     "ts-jest": "^27.1.4",
     "tsconfig": "workspace:*",
-    "typescript": "^4.7.3"
+    "typescript": "4.7.4"
```
mmiszy, line 96: TypeScript doesn't follow SemVer so I removed `^`

apps/saleor-app-checkout/backend/payments/providers/adyen/createPayment.ts
```diff
@@ -27,48 +27,48 @@ export const createAdyenPayment = async (data: OrderFragment, redirectUrl: strin
 
   const checkout = new CheckoutAPI(client);
 
-  const total = data.total.gross;
+  const total = order.total.gross;
 
   const { url, id } = await checkout.paymentLinks({
```
bmigirl, line 32: Maybe we could move this boilerplate to a helper function to make this more easily readable?

apps/saleor-app-checkout/backend/payments/providers/stripe/createPayment.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export const createStripePayment = async ({
+  order,
+  redirectUrl,
+  appUrl,
+  method,
+}: CreatePaymentData): Promise<CreatePaymentResult> => {
+  const stripeClient = await getStripeClient();
+
+  const stripeCheckoutCustomer = await stripeClient.customers.create({
```
bmigirl, line 16: Same as above?

apps/saleor-app-checkout/backend/payments/providers/stripe/createPayment.ts
```diff
@@ -0,0 +1,136 @@
...
+};
+
+const saleorPaymentMethodIdToStripePaymentMethodId = (
+  paymentMethodId: PaymentMethodID
+): Stripe.Checkout.SessionCreateParams.PaymentMethodType | null => {
+  switch (paymentMethodId) {
+    case "creditCard":
+      return "card";
+    case "applePay":
+      // @todo ?
```
bmigirl, line 130: These will be solved in another pr or..?

apps/saleor-app-checkout/backend/payments/providers/stripe/stripeClient.ts
```diff
@@ -0,0 +1,26 @@
+import { getPrivateSettings } from "@/saleor-app-checkout/backend/configuration/settings";
+import { envVars } from "@/saleor-app-checkout/constants";
+import Stripe from "stripe";
+import invariant from "ts-invariant";
+
+export async function getStripeClient() {
+  const { secretKey } = await getStripeSecrets();
+  const stripeClient = new Stripe(secretKey, { apiVersion: "2022-08-01" });
```
bmigirl, line 8: Can we leave a comment here explaining this hardcoded date?

packages/checkout-storefront/src/sections/OrderInfo/PaymentSection.tsx
```diff
@@ -7,17 +7,23 @@ import { getOrderPaymentStatus } from "@/checkout-storefront/fetch";
...
   const { loading: orderPayLoading, orderPay } = usePay();
+  const { env } = useAppConfig();
+
   const [{ data: paymentData, loading: paymentStatusLoading }] = useFetch(getOrderPaymentStatus, {
-    args: { orderId },
+    args: { orderId, checkoutApiUrl: env.checkoutApiUrl },
   });
   const formatMessage = useFormattedMessages();
 
+  console.log({ paymentData, paymentStatusLoading });
```
bmigirl, line 21: this stays on purpose?

bmigirl (approved):

apps/saleor-app-checkout/backend/payments/providers/stripe/createPayment.ts
mmiszy: I don't know yet how to tackle this. It doesn't look like Stripe currently offers a way to do Apple Pay with the payment flow we're using.

apps/saleor-app-checkout/backend/payments/providers/stripe/stripeClient.ts
mmiszy: It's a required setting for stripe. As of now, the only correct value is `2022-08-01`. Literally no other value is accepted.

packages/checkout-storefront/src/sections/OrderInfo/PaymentSection.tsx
mmiszy: Nah, I was debugging something. Thanks!

apps/saleor-app-checkout/backend/payments/providers/stripe/createPayment.ts
bmigirl: So maybe let's create a ticket for that in the project and link the ticket in the comment there? That way in half a year when we see that todo we'll know the reason this wasn't done in the first place

apps/saleor-app-checkout/backend/payments/providers/stripe/stripeClient.ts
bmigirl: What I mean is it looks suspicious being a hardcoded date string, a comment would be helpful to know it's there on purpose 😅 and not e.g a development leftover or something

apps/saleor-app-checkout/config/fields/fields.ts
```diff
@@ -85,6 +87,24 @@ const adyenPaymentProvider: Omit<PaymentProviderSettings<"adyen">, "label">[] =
...
+    type: "string",
+    encrypt: false,
+  },
+  {
+    id: "secretKey",
+    type: "string",
+    encrypt: true,
+  },
+  {
+    id: "webhookSecret",
```
korycins, line 102: You already have secret and public key. Can't we use them to subscribe to Stripe's events? (Also, user doesn't know which event should subscribe) (The webhook creation would be also the way to validate the provided secret&public key data).  Maybe it is not something for MVP, but definitely, it is a must-have for future releases (configuration PaymentPlugin on Saleor side & then on the payment provider was something that caused a lot of issues).

mmiszy: I guess we could create it from the codebase but I'll need to find the best place to do it then :D

apps/saleor-app-checkout/backend/payments/providers/stripe/createPayment.ts
mmiszy: https://github.com/saleor/react-storefront/issues/390

apps/saleor-app-checkout/config/fields/fields.ts
korycins: In Stripe Plugin, we were doing this when the staff user was activating the plugin. I assume that you could do this in a similar way, right?
We store `webhookSecretKey` but we don't show it in plugin configuration form. Instead of this, we show an ID of the webhook instance created by the plugin on Stripe side.
Post MVP sound good, but I would recommend creating at least the task for this

mmiszy: https://github.com/saleor/react-storefront/issues/392

bmigirl (dismissed):

bmigirl (approved):

korycins (approved):

PR #352, opened 2022-08-04
Refund payments in Adyen from Saleor notifications

witoszekdev:
This PR adds Adyen refund handling initiated in Saleor.

- Handle `TRANSACTION_ACTION_REQUEST` events for Adyen transactions
- Add integration tests for webhook handler
- Check if webhook event for specified transaction was already processed to avoid duplicate refunds

Depends on: https://github.com/saleor/react-storefront/pull/321

TODO
- [x] Check if refund exists for a transaction in Saleor (use new transaction query)
- [x] Add integration tests for transaction-action-update webhook endpoint
    - [x] Refund is created
    - [x] Refund is already created
 - ~~Handle VOID events~~ Created separate tickets:
  - https://github.com/saleor/react-storefront/issues/362
  - https://github.com/saleor/react-storefront/issues/361


witoszekdev:
I had to temporarly disable tests for checking if refunds are created in Mollie and Adyen due to issues with Polly.js.

We need to use the node interceptor instead of fetch in Polly.js, but to do that we need to ditch MSW as well. I've made a PR with attempt to fix this some time ago: https://github.com/saleor/react-storefront/pull/325

mmiszy (dismissed):

szczecha (dismissed):

mmiszy (approved):

PR #360, opened 2022-08-05
Replace Cloudfront with Vercel rewrites for demo.saleor.io

witoszekdev:
This PR adds Vercel rewrites instead of Cloudfront for making requests to Saleor Cloud (e.g. `/graphql` route)


apps/storefront/vercel.json
```diff
@@ -0,0 +1,43 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "redirects": [
+    {
+      "source": "/grpahql",
```
mmiszy, line 5: Is this typo on purpose?

witoszekdev: no 🤦, fixed

witoszekdev:
Due to issues with infinite redirects (`/graphql/` -> `/graphql` -> `/graphql/` -> etc.) caused by default Vercel confgiuration, we'll keep on using our custom CloudFront for the demo.

witoszekdev:
let's try this again - with Next.js config instead

apps/storefront/vercel.json
```diff
@@ -0,0 +1,39 @@
+{
```
mmiszy, line 1: Does it work fine with the async rewrites() { we have in next.config.js?
Can we use one of them and not both?

witoszekdev: I think we can use both, as `https://dev-demo.saleor.io/checkout/` works fine. But maybe we should actually just use the Next.js redirects. I'll give it a try :)

witoszekdev: It works fine 👍

mmiszy (approved):

PR #379, opened 2022-08-10
Update Vercel deployment guide

artursmet:
The current guide for Vercel deployment contains outdated build commands. I've updated them to their proper versions I suppose.

However it would be great to have more complete guide on how to deploy the entire storefront + checkout stack from start to the end

### DIFF
--- a/docs/vercel.md
+++ b/docs/vercel.md
@@ -42,7 +42,7 @@ On the configuration page:
 - Override the build command to:
 
 ```bash
-cd ../.. && pnpm run build:saleor-app-checkout
+cd ../.. && pnpm run build --filter=saleor-app-checkout
 ```
 
 - Add environment variables:
@@ -198,7 +198,7 @@ On the configuration page:
 - Override the build command to:
 
 ```bash
-cd ../.. && pnpm run build:checkout
+cd ../.. && pnpm run build --filter=checkout
 ```
 
 - _Optional_: customise [environment variables](../apps/checkout/README.md#env-variables):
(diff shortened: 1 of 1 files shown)


docs/vercel.md
```diff
@@ -42,7 +42,7 @@ On the configuration page:
 - Override the build command to:
 
 ```bash
-cd ../.. && pnpm run build:saleor-app-checkout
+cd ../.. && pnpm run build --filter=saleor-app-checkout
```
lkostrowski, line 45: @mmiszy since this is something our users must do, maybe we can create some aliases in package.json:
```
"build:checkout": "pnpm run build --filter=saleor-app-checkout"
```

witoszekdev: this should be:

```suggestion
cd ../.. && pnpm run build --filter=saleor-app-checkout...
```

without `...` we wouldn't build dependencies of saleor-app-checkout (for example `ui-kit`) and instead use cached version 🙈

https://turborepo.org/docs/core-concepts/filtering#include-dependencies-of-matched-packages

We already had issues with Turborepo cache, it's really nasty thing to debug :/

mmiszy: > @mmiszy since this is something our users must do, maybe we can create some aliases in package.json:
> 
> ```
> "build:checkout": "pnpm run build --filter=saleor-app-checkout"
> ```

Do we add a separate command for every package in the repo?

witoszekdev: @mmiszy I think commands for apps would be enough (storefront, saleor-app-checkout, checkout)

mmiszy: Okay, I can take care of this.

lkostrowski: @mmiszy I only mean root level scripts that can hide turborepo details, as an extra sugar

mmiszy:
Superseded by https://github.com/saleor/react-storefront/pull/391

PR #411, opened 2022-08-23
Change the way storefront passes locale and channel into checkout url

bmigirl:


Cloud11PL (approved):
Makes way more sense 👍

mmiszy (approved):

PR #421, opened 2022-08-25
Remove fragment from GraphQL subscription due to validation issues

witoszekdev:
This PR fixes an issue with app registration caused by Saleor's subscription query validation issues.

It removes fragment from the subscription.


witoszekdev:
Replaced by #423

PR #423, opened 2022-08-25
Add webhooks to manifest

mmiszy:


apps/saleor-app-checkout/pages/api/register.ts
```diff
@@ -1,22 +1,9 @@
...
-  CreateWebhooksMutation,
-  CreateWebhooksMutationVariables,
-  TransactionActionRequestSubscriptionDocument,
-} from "@/saleor-app-checkout/graphql";
-import { getBaseUrl } from "@/saleor-app-checkout/backend/utils";
-import { SALEOR_WEBHOOK_TRANSACTION_ENDPOINT } from "./webhooks/saleor/transaction-action-request";
 import { getAppDomain, setAuthToken } from "@/saleor-app-checkout/backend/environment";
-import urlJoin from "url-join";
 
+// eslint-disable-next-line require-await
```
witoszekdev, line 6: Hmmm... why do we want to disable that rule? If the route doesn't have any `await` then I think we can remove `async`?

witoszekdev (commented):
Really nice... honestly I don't know why I haven't done that sooner 😅

korycins (approved):

PR #427, opened 2022-08-29
Setup jest in checkout

bmigirl:
Also remove some code that actually wasn't is use anymore


packages/checkout-storefront/src/sections/Addresses/utils.ts
```diff
@@ -92,3 +93,41 @@ export const getAddressVlidationRulesVariables = (
...
+  "streetAddress1",
+  "streetAddress2",
+  "city",
+  "postalCode",
+  "cityArea",
+  "countryArea",
+];
+
+// api doesn't order the fields but we want to
+export const getOrderedAddressFields = (addressFields: AddressField[] = []): AddressField[] => {
```
bmigirl, line 111: Threw in something extra for you @mmiszy 😃

mmiszy: 😆

mmiszy (approved):
Well done, looking good

packages/checkout-storefront/package.json
```diff
@@ -11,7 +11,9 @@
     "generate": "DOTENV_CONFIG_PATH=.env graphql-codegen -r env-vars --config codegen.yml",
     "check-types": "tsc --noEmit",
     "lint": "eslint --max-warnings=0 \"src/**/*.@(tsx|ts|jsx|js)\" --fix",
-    "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist"
+    "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist",
+    "test": "jest",
+    "test-dev": "jest --watch"
```
witoszekdev, line 16: Can we use that instead? I think we have this pattern in main `package.json`, so might as well use it here as well :P
```suggestion
    "test:dev": "jest --watch"
```

witoszekdev (commented):
really nice 🥳

packages/checkout-storefront/package.json
bmigirl: Okay

mmiszy (approved):

PR #428, opened 2022-08-29
Some eslint fixes

mmiszy:


apps/saleor-app-checkout/backend/payments/updateTransaction.ts
```diff
@@ -6,7 +6,7 @@ import {
 } from "@/saleor-app-checkout/graphql";
 
 export const updateTransaction = async (args: TransactionUpdateMutationVariables) => {
-  const { data, error } = await getClient()
+  const { data, error: _error } = await getClient()
```
witoszekdev, line 9: `_error` is unused, I think we can remove this

btw. does the new ESLint config ignore `no-unsed-vars` rule when the var starts with `_`?

apps/saleor-app-checkout/package.json
```diff
@@ -9,6 +9,7 @@
     "build": "npm run i18n && next build",
     "start": "next start -p 3001",
     "lint": "next lint",
+    "lint:staged": "next lint --fix",
```
witoszekdev, line 12: Why are we using `--fix` in `lint:staged` in saleor-app-checkout and not inside `lint` like in other packages?

For example: [`packages/checkout-common/package.json`](https://github.com/saleor/react-storefront/pull/428/files#diff-e0d1fd5d7fc07043ea55916c748c9948b090bb0c052116e3f3759cb11ce3e27f)

packages/checkout-storefront/.eslintrc.js
```diff
@@ -1,10 +1,7 @@
 module.exports = {
-  parser: "@typescript-eslint/parser",
+  root: true,
+  extends: ["checkout"],
```
witoszekdev, line 3: Should each package that's not using Next.js extend this config? From what I see `eslint-config-checkout` uses now `next` configuration 🤔

packages/eslint-config-checkout/index.js
```diff
@@ -1,30 +1,75 @@
...
+
+    // we allow empty interfaces
+    "no-empty-pattern": "off",
+    "@typescript-eslint/no-empty-interface": "off",
+
+    // we allow empty functions
+    "@typescript-eslint/no-empty-function": "off",
+
+    // we sometimes use async functions that don't await anything
+    "@typescript-eslint/require-await": "off",
```
witoszekdev, line 50: Offf... I really don't like this. We've specifically added this rule due to issues in API routes in Next.js - sometimes forgetting `await` caused unexpected issues on production that were not visible in development.

I know that this can be annoying (you need to add either `void` or `await`) but IMHO this is a worthy tradeoff. Since API routes are main use case, we could enable it only for them? 🤔

turbo.json
```diff
@@ -48,18 +48,31 @@
...
+    "test": {
       "outputs": []
     },
     "dev": {
       "cache": false
     }
   },
-  "globalDependencies": [".env", ".env.local"]
+  "globalDependencies": [
+    "pnp-lock.yaml",
```
witoszekdev, line 73: Typo

```suggestion
    "pnpm-lock.yaml",
```

turbo.json
```diff
@@ -48,18 +48,31 @@
...
+    ".env",
+    ".env.local",
+    "packages/eslint-config-checkout/index.js"
```
witoszekdev, line 76: Shouldn't this be a dependency of `lint` command only? 🤔

packages/eslint-config-checkout/index.js
```diff
@@ -1,30 +1,75 @@
...
+    project: ["tsconfig.json"],
   },
-  plugins: ["@typescript-eslint"],
-  extends: ["prettier"],
   settings: {
     next: {
       rootDir: ["apps/*/", "packages/*/"],
     },
+    react: {
+      version: "999.999.999",
```
witoszekdev, line 19: Wait what? Why do we need to specify this version? 😅

witoszekdev (changes requested):
Most of the stuff are just commends - except the one about `require-await` - this really has bitten me in the past, so I would prefer to keep using this rule

apps/saleor-app-checkout/package.json
mmiszy: We should be doing `--fix` everywhere

packages/checkout-storefront/.eslintrc.js
mmiszy: Yeah, that's not a problem.

packages/eslint-config-checkout/index.js
mmiszy: https://github.com/jsx-eslint/eslint-plugin-react/issues/1981
I'll add a comment xD

turbo.json
mmiszy: Oh right, yes.

packages/eslint-config-checkout/index.js
mmiszy: Update: seems `"version": "detect"` now works. Not sure if that's related to some changes in eslint or what but thanks for making me check again.

turbo.json
mmiszy: Oh no, not right :D This file is a global dependency of all packages in the repo. Without this line, you'd need to run `--force` to force turborepo to run linter again when any rules inside `packages/eslint-config-checkout/index.js` change.

packages/eslint-config-checkout/index.js
mmiszy: I think you're talking about another rule – `no-misused-promises` or `no-floating-promises` maybe?

`require-await` is actually about ensuring you only add `async` to functions that actually use `await` and remove `async` when functions don't use `await`. Which, in my opinion, is pointless – we often want `async` functions which don't use `await` – we just want to make sure they return a Promise. I don't see any value in this rule apart from **maybe** very minor (negligible) performance hit for using `async` where sync function would be enough.

mmiszy: I'll also enable [`"@typescript-eslint/return-await": ["error", "in-try-catch"],`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/return-await.md)

apps/storefront/components/checkout/SavedAddressSelectionList.tsx
```diff
@@ -29,9 +29,9 @@ export function SavedAddressSelectionList({
 
   if (addresses.length === 0) return null;
 
-  const onSelectSavedAddress = (address: AddressDetailsFragment) => {
+  const onSelectSavedAddress = async (address: AddressDetailsFragment) => {
     setSelectedSavedAddress(address);
-    updateAddressMutation({
+    const _errors = await updateAddressMutation({
```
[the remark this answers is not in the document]
mmiszy: I want us to come back to these places and actually handle errors… :D I'll add some `@todo` comments.

apps/saleor-app-checkout/backend/payments/updateTransaction.ts
mmiszy: Yes

packages/checkout-storefront/src/providers/BillingSameAsShippingProvider.tsx
```diff
@@ -12,7 +12,7 @@ interface BillingSameAsShippingContextConsumerProps {
 export const [useBillingSameAsShipping, Provider] =
   createSafeContext<BillingSameAsShippingContextConsumerProps>();
 
-export const BillingSameAsShippingProvider: React.FC<PropsWithChildren<{}>> = ({ children }) => {
+export const BillingSameAsShippingProvider: React.FC<{ children: ReactNode }> = ({ children }) => {
```
bmigirl, line 15: We have type `children` defined in globalTypes

mmiszy: Yes but it defines `children` as either `ReactNode` or `ReactNode[]`. That's not what we want in most cases.

bmigirl: Yeah but it's that we want in this particular component 😅

bmigirl (approved):

PR #433, opened 2022-08-30
refactor: docs structure

peelar:
## What is the problem?
While getting familiar with this repository, I tried to utilize its documentation, and I noticed what I think are its weaknesses. They all fall down into several categories:
- missing context (e.g. about why you need to tunnel an application)
- a disorder in structure (e.g. there is a section "Deployment" that does not feature any information about the actual deployment process, that information is in a different section)
- too much noise in the top-level README

### Solution

My idea to solve the problems above is to limit the role of the root level README to:
- present the outline of what the repository is
- give instructions on how to get started with the repository without going into details
- redirect to other pages for details

### A problem for another day

Another thing I noticed is that when you follow the current README and try to access the applications on all the different ports, most of them end up "not working". They either display a blank page (`saleor-app-checkout`) or an error page (missing token in `checkout-storefront`). You don't know if you did something wrong, or if this is intentional. I already spoke to @witoszekdev about a potential solution.

## What did I do to solve it?
**The best way to review it is to switch to the branch the PR is built for, and go to `README.md`.**

**Before:**
```
- README.md
- /docs
--  ...
```

**After:**
```
- README.md
- /docs
-- ...
-- development.md ➕
-- /payment ➕
---- index.md ➕
---- mollie.md ➕
---- adien.md ➕
```

- Moved the "payment providers" sections to separate files in a new folder. The folder is called `payment`, and the files are: `payment/index.md`, `payment/mollie.md`, and `payment/adien.md`.
- Created a `docs/development.md` file where I moved more detailed instructions about the development. What's left in the `README.md` is how to install the repository and boot it up.
- I moved some sections around in the `docs/development.md` to maintain a clearer structure.
- I described some context around the tunneling situation in the `saleor-app-checkout` section and the checkout id situation in the `checkout-storefront` section.
- Did some more proofreading, changed some copies, and added some 🌈 **emojis** 🌈.

## What's next?
- I left a comment in the `saleor-app-checkout` section of the `docs/development.md` document. There is one command that felt kind of out of place or perhaps incorrect. Please tell me what to do with it (either delete it or move it somewhere).
- When I first started working with README, I got quickly overwhelmed with it. The actions I took upon it were quite drastic, so if I moved something from the top-level README that you think belongs there, please let me know.
- I spoke to @witoszekdev about the future of `react-storefront` documentation and he suggested we may be already at a point where keeping documentation next to the code is unsustainable. I see two solutions, one temporary and one permanent:
1. (temporary) Provide the "Table of contents" section in the top level README that references all the main sections in all the files that people may be looking for.
2. (permanent) Move the docs to a separate app.


docs/development.md
```diff
@@ -0,0 +1,181 @@
...
+
+You can also start a specific app by running:
+
+```bash
+pnpm dev --filter=saleor-app-checkout
+```
+
+#### [`saleor-app-checkout`](../apps/saleor-app-checkout/README.md)
+
+First, you need to change the default environment variables. Create `.env.local` file in `apps/saleor-app-checkout`, as described [here](../apps/saleor-app-checkout/README.md#env-variables).
```
witoszekdev, line 57: This one is outdated - you can create a single `.env.local` file at the root of monorepo.

I think we should put the "Env vars" paragraph here instead. In practice, you only need to change the Saleor endpoint in `SALEOR_API_URL`. The rest can stay the same for development.

docs/development.md
```diff
@@ -0,0 +1,181 @@
...
+
+> Note: The process needs to be running in the background. Please don't kill it during the development.
+
+<!-- is the bit below correct? you have to filter something to run it for all the apps? -->
+<!-- 
+To run the development server for all the apps, use the following command:
+
+```
+pnpm run dev --filter=saleor-app-checkout...
+```
```
witoszekdev, line 77: No this bit is not correct, you *can* use filter to run only `saleor-app-checkout` + it's dependencies but most of the time you want to run all dev servers. We can drop the `--filter` part.

docs/development.md
```diff
@@ -0,0 +1,181 @@
...
+ -->
+#### [`checkout-storefront`](../packages/checkout-storefront/README.md)
+
+The URL structure for `checkout-storefront` is `localhost:3002?checkout=<ID>`, which means it requires a checkout id to work correctly. You don't have to supply it manually; it will be provided when transitioning from the cart to checkout.
```
witoszekdev, line 81: This bit is also outdated, we now support more than just `checkout` in query params.

Here's an example from staging-demo.saleor.io:

`?locale=pl-PL&checkout=Q2hlY2tvdXQ6ZTcwZjBlYzgtOWMyMS00Y2FkLWE4YzktZWQ3ODcwYjNhN2I5&channel=channel-pln`

The query parameters are:
- `locale` - locale, passed from storefront to checkout, will be used to display messages in correct language
- `checkout` - checkout id (nothing changed here)
- `channel` - channel slug for which we've created the checkout for

@bmigirl Correct me if I'm wrong:
Only `checkout` param is required

docs/development.md
```diff
@@ -0,0 +1,181 @@
...
+  https://my-env.eu.saleor.cloud/graphql/
+  ```
+
+  > To run sandbox Saleor environment in [Saleor Cloud](https://cloud.saleor.io/) use this command:
+  >
+  > ```bash
+  > npx saleor project create && npx saleor environment create
+  > ```
+  >
+  > You can also run Saleor locally. See [Saleor docs](https://docs.saleor.io/docs/3.x/developer/installation) for more instructions
```
witoszekdev, line 148: I think we can drop this bit? Since we're suggesting to use `saleor storefront create` in the main README.

We can leave the link to docs about installation though

docs/development.md
```diff
@@ -0,0 +1,181 @@
...
+
+- `CHECKOUT_APP_URL` — URL of deployed Saleor App Checkout
+
+  Example:
+
+  ```
+  https://saleor-app-checkout.vercel.app
+  ```
+
+  > See [guide below](#vercel) on how to deploy the Saleor App Checkout
```
witoszekdev, line 158: This one is not required for development - it's used to deploy to production, and it's described in `docs/vercel.md` - I think we can remove that

docs/payment/index.md
```diff
@@ -0,0 +1,34 @@
...
+
+<a href="https://www.mollie.com/en">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="./docs/logos/mollie_light.svg">
+    <source media="(prefers-color-scheme: light)" srcset="./docs/logos/mollie_dark.svg">
+    <img alt="Mollie" src="./docs/logos/mollie_dark.svg">
+  </picture>
+</a>
+
+<br>
```
witoszekdev, line 13: I think we can leave that part in main README so that our users know which payment gateways we support out of the box.

@mmiszy We need to add Stripe here as well :D

docs/development.md
```diff
@@ -0,0 +1,181 @@
...
+
+Then, to develop `saleor-app-checkout` app locally, you need to create a tunnel for it. The tunnel will enable you to display it within your Saleor Dashboard.
+
+If you want to read more about Saleor Apps architecture, see [the documentation](https://docs.saleor.io/docs/3.x/developer/extending/apps/key-concepts).
+
+To create a tunnel, run:
+
+```bash
+cd apps/saleor-app-checkout && npx saleor app tunnel 3001
+```
```
witoszekdev, line 67: We duplicate the instructions we have in `apps/saleor-app-checkout/README.md` - maybe we should just link there?

witoszekdev (commented):
I like the change to move more docs out of main README (it was starting to get a bit messy). Great work 👏

I'd like to improve the `docs/development` though - we don't have to do this right now but here are the things missing for new developers who would like to work on this project:
- How to setup env vars used for tests
- Add a note that Vercel previews don't work automatically together - env vars need to be updated in Vercel and the project must be re-deployed
- How to contribute to the project (fork + PR)
- What's the difference between `app/checkout` and `packages/checkout-storefront` (CC @mmiszy)

In deployment guide:
- How redirects work on Vercel (storefront uses Next.js rewrites so that checkout and saleor-app-checkout are on the same domain)
- How to configure error monitoring via Stripe (@mmiszy added that recently)
- How to use CLI to create the project on Vercel (@zaiste showed recently that's something they're working on)
- How to set `CLOUD_DEPLOYMENT_URL` env var - tbh I dunno how our users are supposed to set this up, is this something we need only for demo? @mmiszy

docs/development.md
bmigirl: All of these are required for prod, assuming we make storefronts just point to proper urls. For dev I'd say the assumption is that it's best to supply all for best performance, but we do have some hardcoded defaults just so the code doesn't break in case you forget. That's the best answer I can give

README.md
```diff
@@ -27,14 +27,14 @@
...
-**easily customizable**:
+🛠️ **easily customizable**:
 TailwindCSS can be easily extended and tweaked, or completely replaced with your favorite CSS solution.
 
-**works out-of-the-box**:
-Pre-configured tools for DX.
+👌 **works out-of-the-box**:
+Pre-configured tools built with DX in mind.
 
 ## Stack
```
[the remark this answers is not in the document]
peelar: I thought quite a lot about it and decided to skip this information in README, at least in the context of "bragging" about the core qualities of Storefront (like here).

I understand that both pnpm and Turborepo are cool technologies that may catch people's attention, but these are just tools we used to develop the app that may even change over time. I would go as far as removing the "Stack" section from the README because I think while making the first impression, we should focus on what problems this repository solves, not how.

docs/development.md
peelar: So there is nothing you need to do with env vars specifically for `saleor-app-checkout` to work? It's a general instruction for all the apps?

docs/development.md
peelar: I moved the instructions to `packages/checkout-storefront/README.md`, as suggested with `apps/saleor-app-checkout`.

peelar:
@bmigirl @witoszekdev 

I applied all of your suggestions, besides the one I left unresolved. I would like to deliver it and write down the rest of the suggestions for the next iteration of docs improvements.

README.md
```diff
@@ -27,14 +27,17 @@
...
 ## Motivation
 
-**modern & fast**:
-The project focuses on best practices for e-commerce, like: SSR & ISR, image optimization.
+🏎️ **modern & fast**:
+The project utilizes all the best practices for e-commerce, like SSR, ISR, and image optimization.
 
-**easily customizable**:
+💳 **integrated payments**: 
+Reduce the purchase friction by taking advantage of integrations with modern payment providers such as Adyen or Mollie.
```
witoszekdev, line 34:
```suggestion
Reduce the purchase friction by taking advantage of integrations with modern payment providers such as Adyen, Mollie or Stripe.
```

README.md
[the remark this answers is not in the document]
witoszekdev: All right... but we should still have this information *somewhere* - it's pretty important part of documentation, usefull when debugging issues or adding new features.

To give you an example: Turborepo caches builds. Caching depends on the configuration stored in repository in `turbo.json`, in there you need to specify what causes to "bust" the cache. It's the file output from `build` command + env variables.

Here's the main source of issues I had so far: you forgot to add that env variable to `turbo.json` and 💥 you are wondering for hours why your production didn't change 😅

I wouldn't say it's **bragging** (maybe a litte bit 😎), no less than *Motivation* paragraph.

We also shouldn't assume people are familiar with those technologies - sure you can look stuff up from `package.json`, but what's that `app` and `packages` folders doing there... that's harded to figure out if you didn't have experience with monorepos.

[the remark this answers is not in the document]
peelar: Ah, yes. I see what you mean. I definitely agree that we need to mention Turborepo, I just noticed I completely removed it, even from `docs/development.md`. My main point was that I don't think `README.md` should be the place where we focus on the technologies.

So what I suggest is the following:
1. I will add `Turborepo` to the "Stack" section in `README.md`.
2. I will cover it in greater detail in `docs/development.md`. 

What do you think?

peelar:
@witoszekdev for some reason I can't reply to your comment about Turborepo directly 🤷.

Ah, yes. I see what you mean. I definitely agree that we need to mention Turborepo, I just noticed I completely removed it, even from `docs/development.md`. My main point was that I don't think `README.md` should be the place where we focus on the technologies.

So what I suggest is the following:
1. I will add `Turborepo` to the "Stack" section in `README.md`.
2. I will cover it in greater detail in `docs/development.md`. 

What do you think?

witoszekdev:
@peelar lgtm ;)

mmiszy (approved):

witoszekdev (approved):

PR #443, opened 2022-09-01
fix: :bug: replace hardcoded default channel

peelar:
My store uses a channel called "main". When I tried to set up a storefront for it using `saleor storefront create`, I didn't see my products, even though I changed the URL from `http://localhost:3000/default-channel/en-US` to `http://localhost:3000/main/en-US`.

I thought this is perhaps due to some environment variable being assigned a default value, so I went to `apps/storefront/.env` and I changed `NEXT_PUBLIC_DEFAULT_CHANNEL=default-channel` to `=main`. That also didn't solve it.

Then I noticed that in `apps/storefront/lib/regions.ts` we have an object called `DEFAULT_CHANNEL` that uses a hardcoded value for the default channel slug. I replaced that with the .env value and that solved it.

### DIFF
--- a/apps/storefront/lib/regions.ts
+++ b/apps/storefront/lib/regions.ts
@@ -1,6 +1,7 @@
 import { GetStaticPropsContext } from "next";
 
 import { LanguageCodeEnum } from "../saleor/api";
+import { DEFAULT_CHANNEL as CHANNEL_SLUG } from "./const";
 
 export const LOCALES = [
   {
@@ -23,7 +24,7 @@ export interface Channel {
 }
 
 export const DEFAULT_CHANNEL: Channel = {
-  slug: "default-channel",
+  slug: CHANNEL_SLUG,
   name: "United States Dollar",
   currencyCode: "USD",
 };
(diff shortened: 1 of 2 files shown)


bmigirl:
Closing this because SF project is a template for people to use as their own storefronts base -> hence the channels configuration needs to be replaced accordingly to channels set in dashboard, rather than adding more env vars. @peelar we might want to improve readme to point that out

witoszekdev:
It *is* a template but we need to do something with that env var - it's not used anywhere, we might as well just use it for configuration (imho it's easier to edit env vars than to look through the project and replacing hardcoded values)

apps/storefront/lib/regions.ts
```diff
@@ -23,7 +24,7 @@ export interface Channel {
 }
 
 export const DEFAULT_CHANNEL: Channel = {
-  slug: "default-channel",
+  slug: CHANNEL_SLUG,
   name: "United States Dollar",
   currencyCode: "USD",
```
krzysztofwolski, line 29: If channel from the env has a different currency, then `name` and `currencyCode` fields will not be valid anymore

krzysztofwolski (changes requested):

apps/storefront/lib/regions.ts
krzysztofwolski: Additional variables for configuring it might solve the issue

krzysztofwolski:
BTW I've seen a few occurrences of "default-channel" string in the checkout part, please check if changing the value for a channel in storefront wont break other apps

Issue #364, opened 2022-08-08
Fix checkout query always returning loading: false


bmigirl:
Currently checkout query always returns loading `false` - we need to fix that, so e.g summary can properly display loaders when updating quantity etc. Also we now have a bug where checkout finalization sometimes fails on total amount mismatch - this is due to the finalization waiting till all the queries and mutations are done, hence checkout returning loading: false makes us sometimes send wrong data to the api

PR #455, opened 2022-09-05
Remove disabling pay button

bmigirl:


mmiszy:
What about #437?

bmigirl:
@mmiszy also closed [#437](https://github.com/saleor/react-storefront/pull/437) because it was an outdated branch without latest fixes

bmigirl:
Also added description to [this issue](https://github.com/saleor/react-storefront/issues/364) which should fix a problem with total amount mismatch that is still happening here sometimes

packages/checkout-storefront/src/hooks/useCheckoutUpdateStateTrigger.ts
```diff
@@ -0,0 +1,17 @@
...
+import { useFormContext } from "react-hook-form";
+
+export const useCheckoutUpdateStateTrigger = (
+  scope: CheckoutUpdateStateScope,
+  updating: boolean
+) => {
+  const { setValue } = useFormContext<CheckoutFormData>();
+
+  useEffect(() => {
+    setValue(`updateState.${scope}`, updating);
```
witoszekdev, line 15: Why are we setting some values in `react-hook-form`? Is it just to trigger some validation updates?

witoszekdev (commented):
I generally feel like we're more fighting with `react-hook-form` than actually benefit from using it. I guess we don't have to do this *right now* but perhaps we should replace it with some other library / write our own hooks?

packages/checkout-storefront/src/sections/Addresses/useAddressFormUtils.ts
```diff
@@ -1,17 +1,46 @@
...
+      (result, fieldValue, fieldName) => {
+        if (!isRequiredField(fieldName as AddressField)) {
+          return result;
+        }
+
+        return !!fieldValue ? result : ([...result, fieldName] as AddressField[]);
+      },
+      [] as AddressField[]
+    );
+  };
```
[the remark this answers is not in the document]
mmiszy: @witoszekdev because `address` is an object but we could do `Object.entries(address).reduce(…)`

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutForm.ts
```diff
@@ -0,0 +1,108 @@
...
+    ...getValues(),
+    paymentProviderId: selectedPaymentProvider as PaymentProviderID,
+    paymentMethodId: selectedPaymentMethod as PaymentMethodID,
+  });
+
+  console.log({ loadingCheckout });
+  const hasFinishedApiChanges =
+    !Object.values(methods.watch("updateState")).some((value) => value) && !loadingCheckout;
+
+  // not using form handleSubmit on purpose
```
mmiszy, line 79: Comments should answer the question "Why?". I can see what's going on but I don't understand _why_.

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutFormValidation.ts
```diff
@@ -0,0 +1,143 @@
...
+        showCustomErrors([{ field: path as string, code: type as ErrorCode }]);
+        isValid = false;
+      }
+    }
+
+    if (!checkout.shippingAddress) {
+      showCustomErrors([{ field: "shippingAddress", code: "required" }]);
+    }
+
+    if (!shippingHasAllRequiredFields(checkout.shippingAddress)) {
```
mmiszy, line 97: Looks like we could use some form-splitting at this point, to make `shippingAddress` and `billingAddress` kind of independent forms with their own validation.

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutFormValidation.ts
```diff
@@ -0,0 +1,143 @@
...
+  };
+
+  const validating = watch("validating");
+
+  // needs to be a hook so react doesn't batch this
+  useEffect(() => {
+    if (validating) {
+      setValue("validating", false);
+    }
+  }, [validating]);
```
mmiszy, line 140: What problem are we trying to solve here?

mmiszy (changes requested):

packages/checkout-storefront/src/lib/utils/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from "./utils";
 export * from "./validation";
 export * from "./money";
+export * from "./tests";
```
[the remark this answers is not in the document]
bmigirl: Can you propose something better? I'm not sure myself where this should land so I'd appreciate suggestions 😅

packages/checkout-storefront/src/hooks/useAlerts/useAlerts.tsx
```diff
@@ -48,18 +48,32 @@ function useAlerts(globalScope?: any): any {
...
   };
 
-  const showAlert = ({ scope, ...dataRest }: AlertErrorData, { type }: { type: AlertType }) => {
-    const { message, ...options } = getParsedAlert({ ...dataRest, scope }, type);
-    toast(message, options);
+  const showDefaultAlert = (
+    { scope, ...dataRest }: AlertErrorData,
+    { type }: { type: AlertType } = { type: "error" }
+  ) => {
+    const parsedAlert = getParsedAlert({ ...dataRest, scope }, type);
```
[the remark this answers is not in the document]
bmigirl: Must be leftover, changing

packages/checkout-storefront/src/lib/utils/index.ts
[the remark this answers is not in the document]
mmiszy: Yes, let's remove this line.

We can also move src/lib/utils/tests.ts somewhere else - where we have other test related functions.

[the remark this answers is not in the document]
mmiszy: We have similar functions in `saleor-app-checkout/__tests__/utils` so maybe we can do the same here? `__tests__/utils`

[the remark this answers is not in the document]
bmigirl: ok

packages/checkout-storefront/src/hooks/useCheckoutUpdateStateTrigger.ts
bmigirl: Yeah, it's because checkout finalization now needs to wait till all api changes happen, hence we're have a state in checkout form so it knows whether to wait or not

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutFormValidation.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  return ensureValidCheckout;
```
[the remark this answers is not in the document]
bmigirl: I'm with you @witoszekdev, talked to @mmiszy as well and we all agree it's hacky and react hook form needs a replacement. I added a ticket [here](https://github.com/saleor/react-storefront/issues/475)

packages/checkout-storefront/src/sections/Addresses/useAddressFormUtils.ts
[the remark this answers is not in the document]
witoszekdev: Yes, I meant `entries` not `fromEntries` 🤦🏻‍♂️

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutFormValidation.ts
witoszekdev: @bmigirl I think we can use `flushSync` instead of `useEffect` - https://github.com/reactwg/react-18/discussions/21

bmigirl: Oh yeah, right. Thanks lots, will change

packages/checkout-storefront/src/lib/translations/en-US.json
```diff
@@ -11,6 +11,7 @@
   "account": "Account",
   "signOut": "Sign Out",
   "signIn": "Sign In",
+  "checkoutFinalizeEmailInvalidError": "Please provide valid email before finalizing checkout",
```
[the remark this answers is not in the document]
bmigirl: This is needed because we generate those keys based on mutation name `checkoutFinalize` + field name `Email` + error type `Invalid` + `Error` -> and we don't type them anywhere in the code manually. So the fact these two are the same right now is only because of the fact that I didn't come up with a better messages, but they're needed

witoszekdev (dismissed):

szczecha (approved):

packages/checkout-storefront/jest.config.js
```diff
@@ -4,6 +4,7 @@ const { compilerOptions } = requireJSON("./tsconfig.json");
 
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
 module.exports = {
+  testMatch: ["**/**/*.test.ts"],
```
witoszekdev, line 7: If that's a glob shouldn't this also work?

```suggestion
  testMatch: ["**/*.test.ts"],
```

packages/checkout-storefront/package.json
```diff
@@ -43,13 +44,14 @@
...
     "react-error-boundary": "^3.1.4",
-    "react-hook-form": "^7.33.1",
+    "react-hook-form": "^7.34.2",
     "react-test-renderer": "^18.2.0",
     "react-toastify": "^9.0.4",
     "tailwindcss": "^3.0.18",
     "ts-invariant": "0.10.3",
     "url-join": "^5.0.0",
     "urql": "^2.1.3",
+    "wonka": "^6.0.0",
```
mmiszy, line 54: Should this be in `devDependencies`?

mmiszy (approved):

packages/checkout-storefront/package.json
bmigirl: It is now 😅

mmiszy (approved):

PR #462, opened 2022-09-06
Suggest using at least pnpm 7.7.1; fix Vercel to use 7.11.0

mmiszy:
pnpm introduced several breaking changes in the lockfile in versions 7.6.0 and 7.7.0.

I've generated this lockfile with 7.7.1 and tested with 7.8.0, 7.9.0 and 7.11.0 – all work fine.
Hence, we suggest using anything above 7.7.0 but pin Vercel to 7.11.0.


apps/checkout/package.json
```diff
@@ -47,5 +47,5 @@
     "node": ">=16.0.0",
     "pnpm": "^7.5.0"
```
witoszekdev, line 48: We should update it just like in main `package.json`:

```suggestion
    "pnpm": ">=7.7.1"
```

apps/storefront/package.json
```diff
@@ -81,5 +81,5 @@
     "node": "^14 || ^16 || ^18",
     "pnpm": "^7.5.0"
```
witoszekdev, line 82:
```suggestion
    "pnpm": ">=7.7.1"
```

packages/ui-kit/package.json
```diff
@@ -79,5 +79,5 @@
     "node": ">=16.0.0",
     "pnpm": "^7.5.0"
```
witoszekdev, line 80:
```suggestion
    "pnpm": ">=7.7.1"
```

witoszekdev (approved):

lkostrowski (approved):
Acknowledged :) @zaiste @2can FYI - please check if this changes something in CLI flow.

CLI should not be aware of the implementation details, so hopefully CLI can read pnpm version (if even has to) from package and consume it. I hope we don't have to keep track on such changes in the future

Issue #466, opened 2022-09-07
"Deploy to Vercel" button doesn't work


lkostrowski:
### Summary

In readme (https://github.com/saleor/react-storefront/tree/canary/apps/storefront#vercel--netlify) there is a button that should automatically clone and deploy storefront to Vercel.

However, this doesn't work. At the moment, it fails on PNPM, but AFAIK it will continue failing, because "button" is constructing a query that doesn't work for monorepo

Here is Vercel reference https://vercel.com/docs/deploy-button

That provides configuration like build & install script and root directory.

### Steps to reproduce

1. Open docs - https://github.com/saleor/react-storefront/tree/canary/apps/storefront#vercel--netlify
2. Click Deploy to Vercel button
3. Proceed with Vercel deployment until build fails

lkostrowski:
@mmiszy This is not closed via #469. This PR enabled next config, but it has to be configured for workspaces. You can try - it fails

mmiszy:
Thanks @lkostrowski

PR #469, opened 2022-09-07
Add vercel.json files to storefront and app-checkout packages

lkostrowski:
This points Vercel to explicitly set Next as framework in deployment context

# Why this is needed

[Deploy button api](https://vercel.com/docs/deploy-button) can't specify framework nor output directory settings. In monorepo, when react-storefront is being spawned, Next creates deployment with framework "Other" which is probably a bug.

It was reported to Vercel, but this local fix is transparent and fixes the problems

It will also be needed to close https://github.com/saleor/react-storefront/issues/466


krzysztofwolski (dismissed):

mmiszy (commented):
Looks good but I think you're missing empty lines at the end of both files?

mmiszy:
Question: where was it reported to Vercel? Do you have a link to the issue so we can track it?

lkostrowski:
@mmiszy 

Added empty lines - but I suggest to add this to eslint/prettier 

Regarding Vercel issue - I reported it from contact form yesterday, ticket is assigned to my account, will let you know what will be there answer...

...but when I think about it, vercel.json *should* be there even if this works. For tooling (CLI etc) the less "details" the better, so if application can store its config in the repository, that's perfect. We can narrow down integration to as small interface as possible

mmiszy:
All files are prettified on commit – you must've skipped the pre-commit git hook @lkostrowski
I just tested this now on another branch and `vercel.json` files were automatically prettified for me when I committed.

lkostrowski:
@mmiszy right, I probably changed it without installing deps, hence no husky too.

Also, I received answer from Vercel, they suggested to use vercel.json, so I assume we are doing right.

lkostrowski:
@saleor/storefront-checkout I added also build and install overwrites to apps. Thanks to that, tooling that deploy storefront and checkout can ignore implementation details and integration should be less flaky

Issue #419, opened 2022-08-25
UX improvements


timuric:
• Use pointers for all clickable elements
• Close edit address view on save
• Use html select
• For checkout without shipping (virtual goods) change copy from "Use shipping address as billing address" to "Add shipping address"
• Change button name to "Add new address"
![image](https://user-images.githubusercontent.com/4006792/186656682-a2884031-1cd5-4aa3-a35b-432773e0f3a1.png)

• Background is blue
![image](https://user-images.githubusercontent.com/4006792/186657013-8f34c4ee-e6d1-4fd7-b8e0-bcbbccd4658b.png)

• Remove title `problemTitle`

• Remove "Remember your password?" 
![image](https://user-images.githubusercontent.com/4006792/187709230-bd510bb2-27ca-4b08-a8b9-fe8cbc35105d.png)

Remove double border
![image](https://user-images.githubusercontent.com/4006792/188658320-35f6f8d5-07b9-452c-9aab-44bfafd6d34c.png)

Issue #417, opened 2022-08-25
Checkout keyboard navigation


mmiszy:
A few bugs in checkout:

1. Links (Sign In / Sign Out etc) are selectable with keyboard but there's no outline which indicates what's actually selected.
2. Checkboxes are selectable with keyboard but there's no visual indicator of that.
3. Dropdowns (country, state) are not selectable with keyboard
4. The State dropdown is looooooooong and not scrollable (you scroll the whole page): https://www.loom.com/share/aeebd5349c9547a39363e7bfcf9abb8d
5. Delivery methods are not selectable with keyboard – they should be input type="radio"
6. Payment methods, same as above ^

mmiszy:
Rel: https://github.com/saleor/react-storefront/issues/366

mmiszy:
Closed in https://github.com/saleor/react-storefront/pull/489

PR #489, opened 2022-09-22
Ux improvements

bmigirl:
Solves #417 and #419


packages/checkout-storefront/src/components/SelectBox/SelectBox.tsx
```diff
@@ -1,34 +1,32 @@
 import clsx from "clsx";
 import { Children, Classes } from "@/checkout-storefront/lib/globalTypes";
-import { FC } from "react";
+import { FC, HTMLAttributes } from "react";
 
-export interface SelectBoxProps extends Classes, Children {
+export interface SelectBoxProps
+  extends Classes,
+    Children,
+    Omit<HTMLAttributes<HTMLInputElement>, "children"> {
```
witoszekdev, line 8: I think this one is more clear, because DOM types look exactly the same (`HTML...`). If we'd use `React.` namespace, it would be more clear that the types came from React
```suggestion
    Omit<HTMLAttributes<React.HTMLInputElement>, "children"> {
```

packages/checkout-storefront/src/sections/DeliveryMethods/DeliveryMethods.tsx
```diff
@@ -44,38 +42,42 @@ export const DeliveryMethods: React.FC<CommonSectionProps> = ({ collapsed }) =>
...
-    }
-
-    setSelectedMethodId(deliveryMethod.id);
-  }, [deliveryMethod]);
-
-  const handleSubmit = async (selectedMethodId: string) => {
-    setSelectedMethodId(selectedMethodId);
-
+  const handleSubmit = async ({ selectedMethodId }: FormData) => {
+    console.log({ selectedMethodId });
```
witoszekdev, line 80:
```suggestion
```

packages/checkout-storefront/src/sections/PaymentSection/PaymentMethods.tsx
```diff
@@ -1,29 +1,86 @@
...
+      return [];
+    }
+
+    return compact(Object.keys(allPaymentMethods)) as AvailablePaymentMethods;
+  };
+
+  const availablePaymentMethods = getParsedPaymentMethods(allPaymentOptions);
+
+  useEffect(() => {
+    if (!loading && allPaymentOptions && !availablePaymentMethods.length) {
```
witoszekdev, line 58: Why do we need `allPaymentOptions` here?

packages/checkout-storefront/src/sections/PaymentSection/PaymentMethods.tsx
```diff
@@ -1,29 +1,86 @@
...
+      throw new Error("No available payment providers");
+    } else if (!loading && availablePaymentMethods.length && !selectedPaymentMethod) {
+      setValue("paymentMethodId", availablePaymentMethods[0] as PaymentMethodID);
+    }
```
witoszekdev, line 62: I think this would make it a little bit more readable
```suggestion
    if (loading) return;
    
    if (!availablePaymentMethods.length) {
        throw new Error("No available payment providers");
    } else if (!selectedPaymetMethod) {
        setValue("paymentMethodId", availablePaymentMethods[0] as PaymentMethodID);
    }
```

packages/checkout-storefront/src/sections/PaymentSection/PaymentMethods.tsx
bmigirl: Actually we don't, removing 👍

witoszekdev (approved):

mmiszy (approved):

PR #494, opened 2022-09-26
Speed up build and dev by getting rid of Babel

### LINKED ISSUE #479: Remove .babelrc from Next apps
Opt-in into using SWC instead of Babel.
Replace formatjs babel plugin with eslint formatjs plugin.

mmiszy:
- [x] get rid of babel
- [x] replace formatjs babel plugin with eslint plugin
- [x] fix eslint configuration to lint all files

|              | **before** | **after** |  **speedup**  | 
|--------------|:----------:|:---------:|:------------|
|    **cache** |   24.69s   |   22.82s  |  ✅ **–7.57%**  | 
| **no-cache** |   80.37s   |   59.59s  | ✅ **–25.86%** | 

Fixes #479


apps/saleor-app-checkout/setupTestsAfterEnv.ts
```diff
@@ -18,6 +18,7 @@ afterEach(() => server.resetHandlers());
 afterEach(() => {
   consoleTypes.forEach((type) => {
     // @ts-expect-error
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
     console[type]?.mockClear?.();
```
witoszekdev, line 22: This is more safe, because we're not making an unsafe `?.()` call. Optional chaining function call is not actually checking if the property is a function before calling it, instead it only checks if the value is not `undefined` or `null`. This means if, for some reason, it's something else (e.g. a string / an object) then this call will throw. I dunno why that's the spec, it makes no sense 🤦🏻‍♂️ 

```suggestion
    typeof console[type]?.mockClear === "function" && console[type].mockClear();
```

apps/saleor-app-checkout/setupTestsBeforeEnv.ts
```diff
@@ -6,6 +6,7 @@ process.env.SALEOR_APP_TOKEN = "TEST";
 // Fix for jose missing TextEncoder error
 // https://github.com/inrupt/solid-client-authn-js/issues/1676
 if (typeof global.TextEncoder === "undefined") {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   const { TextEncoder, TextDecoder } = require("util");
```
witoszekdev, line 10: Could we do this?

```suggestion
  import { TextEncoder, TextDecoder } = require("util");
```

apps/saleor-app-checkout/test-utils.ts
```diff
@@ -89,7 +91,7 @@ const tryParse = (text: string | undefined) => {
 
 export const setupPollyMiddleware = (server: PollyServer) => {
   // Hide sensitive data in headers or in body
-  server.any().on("beforePersist", (_, recording, event) => {
+  server.any().on("beforePersist", (_, recording, _event) => {
```
witoszekdev, line 94: can we just remove it?
```suggestion
  server.any().on("beforePersist", (_, recording) => {
```

witoszekdev (commented):
Yay! 🥳 

image

patrys:
@mmiszy Can we get a before/after?

bmigirl (approved):

mmiszy:
Added benchmarks to the description @patrys

PR #512, opened 2022-10-05
Fix add variant to cart test after removing cart view

szczecha:
Fix test and remove cypress from dependencies in app/storefront


mmiszy (approved):

mmiszy:
You need to run `pnpm i` to update the lockfile @szczecha

mmiszy (approved):

cypress/support/pages/product-page.ts
```diff
@@ -12,47 +12,21 @@ export function openProductPage(product) {
...
-            .invoke("text")
-            .then((selectedVariantName) => {
-              cy.get(PRODUCT_ELEMENTS.variant)
-                .first()
-                .click()
-                .get(PRODUCT_ELEMENTS.addToCartButton)
-                .click()
-                .url()
-                .should("include", "/cart")
-                .get(CART.cartVariant)
```
karola312, line 34: Remove CART import it is no longer used

cypress/e2e/addVariantToCart.cy.ts
```diff
@@ -1,23 +1,22 @@
...
-    cy.visit("/");
+    cy.visit("/").clearLocalStorage();
   });
 
   it("should select a variant and add to the cart SRS_0202", () => {
     const product = productsToSearch.productWithVariants;
 
     openProductPage(product);
-    addItemToCart(PRODUCT_ELEMENTS.productName);
+    addItemToCart();
```
karola312, line 13: It would be nice if at the end we check the cart icon - there should be a number of items in the cart

szczecha: In the future, there will be a mini cart and then the test will be checking if that product is in the cart.

PR #522, opened 2022-10-10
Fix Stripe charged bug

### LINKED ISSUE #521: Stripe order confirmation bug
### Discussed in https://github.com/saleor/react-storefront/discussions/492

<div type='discussions-op-text'>

<sup>Originally posted by **jTiKey** September 24, 2022</sup>
I set up the system and I'm getting stripe webhook events, but the backend doesn't seem to "mark as paid" orders, even if they are paid. The order history shows "checkout.session.completed".

The Plugin stripe webhook isn't usable, since the payment intent is not created by the checkout app.

I guess this is not written yet? I'll have to rewrite the webhook logic manually?</div>

mmiszy:
Fixes #521


apps/saleor-app-checkout/backend/payments/providers/stripe/webhookHandler.ts
```diff
@@ -91,10 +91,11 @@ export const checkoutSessionToTransactionCreateMutationVariables = async (
     }
 
     const getAmount = getTransactionAmountGetter({
+      // in Stripe, authorized means that the payment is authorized and charged
       authorized: getSaleorAmountFromInteger(charge.amount),
+      charged: getSaleorAmountFromInteger(charge.amount),
```
witoszekdev, line 96: This could be a bit problematic because transaction amounts in different states add up in Saleor.

What I mean is that the total transaction amount will be `authorized + charged` - what we should do is set it either as authorized **or** charged not both at the same time, because it will mess up calculating remaining amount to pay in order (the order will be overpaid). https://docs.saleor.io/docs/3.x/developer/checkout#updating-transaction

mmiszy: Good catch! I actually copied it from another PR (multi tenant) and I didn't verify. My fault.

PR #523, opened 2022-10-11
Fix checkout not waiting for refetching checkout after mutations

bmigirl:


packages/checkout-storefront/src/hooks/useCheckoutUpdateStateTrigger.ts
```diff
@@ -12,8 +12,18 @@ export const useCheckoutUpdateStateTrigger = (
...
-    setValue(`updateState.${scope}`, updating);
+    if (updating) {
+      setValue(stateKey, updating);
+    } else {
+      // temporary solution to make sure we wait for checkout fetch
+      // that'd happen soon after any mutations. will be fixed
+      // once we implement advanced caching for urql
+      setTimeout(() => {
+        setValue(stateKey, updating);
+      }, 2000);
```
mmiszy, line 26: Why 2000? Is there a special meaning to this number?

bmigirl: Not really, it's a smallest timeout that worked

witoszekdev (approved):

PR #528, opened 2022-10-13
Fix Dockerfile

witoszekdev:
This PR is indented to fix Dockerfile to make `react-storefront` work with `saleor-platform`

Related `saleor-platform` PR: https://github.com/saleor/saleor-platform/pull/192

- Exposed a `docker-compose` file that contains services for each app used inside monorepo
- Each app extends a single `Dockerfile.base` and then modifies it's `CMD` that's used to run the app
- `storefront` has nginx configuration for communicating with saleor (it assumes the configuration from `saleor-platform`) and with `saleor-app-checkout`
  - using Docker's service names directly doesn't work properly (connection refused error)

Due to limitations with turborepo + pnpm + Next.js standalone mode the images have quite large size ~2GB (compared to ~200 MB in standalone mode)

For further research:
- [pnpm issues with Docker](https://github.com/pnpm/pnpm/issues/3114)
- [Example project with pnpm store configured](https://github1s.com/gaggle/exploring-the-monorepo/tree/attempt-perfect-docker)
- [Next.js Standalone mode docs](https://nextjs.org/docs/advanced-features/output-file-tracing)
- [Turborepo Docker guide](https://turborepo.org/docs/handbook/deploying-with-docker)
- [Next.js example Dockerfile](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile)


Dockerfile.base
```diff
@@ -0,0 +1,49 @@
...
+ARG SENTRY_ENVIRONMENT
+ENV SENTRY_ENVIRONMENT ${SENTRY_ENVIRONMENT}
+
+ARG SENTRY_RELEASE
+ENV SENTRY_RELEASE ${SENTRY_RELEASE}
+
+ARG VERCEL_URL
+ENV VERCEL_URL ${VERCEL_URL:-localhost:3001}
+
+ENV ENABLE_EXPERIMENTAL_COREPACK 1
```
lkostrowski, line 47: isn't it Vercel related? We control pnpm here

Dockerfile.dev
```diff
@@ -5,15 +5,23 @@ WORKDIR /app
 RUN npm install -g pnpm@7.11.0
```
lkostrowski, line 6: why Dockerfile.dev? It looks like it runs actually production? For dev its most likely to not use Docker at all

witoszekdev: my bad, forgot to delete that file

Dockerfile.base
witoszekdev: I've copied over envs from `.env` file but it looks like that one isn't used anywhere - good catch

apps/saleor-app-checkout/Dockerfile
```diff
@@ -0,0 +1,48 @@
+# Export package.json's in monorepo using turborepo
+# https://turborepo.org/docs/handbook/deploying-with-docker#the-solution
+FROM node:16-slim AS builder
+WORKDIR /app
+
+ARG PNPM_VERSION=7.13.4
```
mmiszy, line 6: Why are we using different pnpm version than in Dockerfile.base?

apps/saleor-app-checkout/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+# -- Install dependencies using pnpm
+
+FROM node:16-slim AS installer
+WORKDIR /app
+RUN npm --no-update-notifier --no-fund --global install pnpm@${PNPM_VERSION}
+
+COPY .gitignore .gitignore
+COPY --from=builder /app/out/json/ .
+COPY ./pnpm-lock.yaml ./pnpm-lock.yaml
+COPY ./.npmrc ./.npmrc
```
mmiszy, line 21: We don't need this I guess

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
+version: "3"
+
+services:
+  saleor-app-checkout:
+    container_name: saleor-app-checkout
+    build:
+      context: .
+      dockerfile: Dockerfile.base
```
mmiszy, line 8: Is this correct?

apps/saleor-app-checkout/Dockerfile
witoszekdev: That's a leftover from trying to build checkout in standalone mode - I've removed it

apps/saleor-app-checkout/Dockerfile
witoszekdev: yes, removed the file

docker-compose.yml
witoszekdev: I'm not sure what you meant here? Both services use the same Dockerfile, they only change the command used to run the service.

`context: .` is so that they are run from the root of monorepo.

`container_name` is arbitrary tbh, we can change it

mmiszy (approved):

PR #529, opened 2022-10-13
Fix iFrame embedding of Checkout

mmiszy:


witoszekdev (approved):

mmiszy:
This is a temporary solution. It reverts `checkout` headers to what they were previously before I added a proxy through the Storefront.

PR #535, opened 2022-10-17
Add changing localization

bmigirl:


packages/checkout-storefront/src/components/AddressForm/AddressForm.tsx
```diff
@@ -42,22 +43,24 @@ export const AddressForm: FC<PropsWithChildren<AddressFormProps>> = ({
     formState: { isDirty },
   } = formProps;
   const formData = watch();
+  const formatMessage = useFormattedMessages();
   const getInputProps = useGetInputProps(formProps, defaultInputOptions);
   const { isAvailable } = useAddressAvailability({ pause: !checkAddressAvailability });
+  const ref = useRef<HTMLFormElement>(null);
```
mmiszy, line 49: Is this used anywhere?

packages/checkout-storefront/src/hooks/useUrlChange.ts
```diff
@@ -0,0 +1,18 @@
+import { getQueryParams, QueryParams } from "@/checkout-storefront/lib/utils";
+import { useEffect } from "react";
+
+export const URL_CHANGED = "urlChangedEvent";
```
mmiszy, line 4: There's no such DOM event

mmiszy (changes requested):

packages/checkout-storefront/src/components/AddressForm/AddressForm.tsx
bmigirl: Just testing if your guard is up

mmiszy (approved):

packages/checkout-storefront/src/sections/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,12 @@
+import { SaleorLogo } from "@/checkout-storefront/images";
+import { getSvgSrc } from "@/checkout-storefront/lib/svgSrc";
+import { LanguageSelect } from "@/checkout-storefront/sections/PageHeader/LanguageSelect";
+
+export const PageHeader = () => {
+  return (
+    <div className="page-header">
+      <img src={getSvgSrc(SaleorLogo)} alt="logo" className="logo" />
```
witoszekdev, line 8:
```suggestion
      <img src={getSvgSrc(SaleorLogo)} alt="" className="logo" />
```

I don't think this one needs an alt at all - if we give it an empty string it will be skipped in a11y tree

PR #536, opened 2022-10-17
Update i18n commands and add Minion translation

mmiszy:


bmigirl (approved):

.gitignore
```diff
@@ -44,4 +44,7 @@ cypress/videos
 
 .idea
 
-.auth_token
\ No newline at end of file
+.auth_token
+
+
```
witoszekdev, line 49: why so many lines? 😅

.husky/pre-commit
```diff
@@ -1,4 +1,8 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
+pnpm run i18n:extract
+git add apps/saleor-app-checkout/content/locales
+git add packages/checkout-storefront/content/locales
```
witoszekdev, line 6: hmm... I'm not a big fan of running long-running commands and adding new files in commits automatically - could we instead check if the locales were extracted on `git push`?

IMHO, we should offload checks as much as possible to CI, so we don't have to wait locally for stuff to complete - this is a huge issue in dashboard, every commit takes ~30 seconds + 2/3 minutes on each push because of slow commands. This is really irritating for small changes

apps/saleor-app-checkout/frontend/hooks/useFormattedMessages.ts
```diff
@@ -1,8 +1,8 @@
 import { useRouter } from "next/router";
 import { useMemo } from "react";
 
-import English from "../content/compiled-locales/en.json";
-import Polish from "../content/compiled-locales/pl.json";
+import English from "../../content/compiled-locales/en.json";
+import Polish from "../../content/compiled-locales/pl.json";
```
witoszekdev, line 5: This will cause to download all locales, which could be a problem if we add more of them - maybe we should import them dynamically?

witoszekdev (changes requested):

apps/saleor-app-checkout/frontend/hooks/useFormattedMessages.ts
mmiszy: Maybe, one day :P

.husky/pre-commit
mmiszy: `check if the locales were extracted` - this would take the same amount of time or even more, and when it fails you'd need to run the command and `git add` manually, and try to commit again. I don't see a point in that if we can automate.

apps/saleor-app-checkout/frontend/hooks/useFormattedMessages.ts
bmigirl: Actually that makes sense, I think we had something like that in dashboard

.husky/pre-commit
mmiszy: But we're not running TS or Eslint on commit @bmigirl so I'm not sure what are you referring to

mmiszy: And you really shouldn't commit with `no-verify` :P

apps/saleor-app-checkout/frontend/hooks/useFormattedMessages.ts
mmiszy: https://github.com/saleor/react-storefront/issues/537

.husky/pre-commit
mmiszy: https://github.com/saleor/react-storefront/issues/538

PR #547, opened 2022-10-19
Fix removing pathname from url after language change

bmigirl:


packages/checkout-storefront/src/lib/utils/utils.ts
```diff
@@ -53,7 +53,7 @@ export const clearUrlAfterPasswordReset = (): void => {
 };
 
 export const replaceUrl = ({
-  url = window.location.origin,
+  url = window.location.origin + window.location.pathname,
```
mmiszy, line 56: I think we can change this to `window.location.href` or `window.location.toString()`
```suggestion
  url = window.location.toString(),
```

mmiszy (changes requested):

mmiszy (approved):

PR #592, opened 2022-10-27
[WIP & POC & Blocked] Make storefront app

lkostrowski:
This PR proves new approach for Storefront:

- Storefront doesnt' require statically build checkout URL to redirect.
- Storefront is a Saleor App
- Storefront in Dashboard can save checkout URL string to AppMetadata
- Storefront in client side can read checkout URL and uses it to redirect

PR is heavily WIP and before code is reviewed, lets discuss proposed approach (urql, Macaw/Material etc) - comments by me left in the code

**Blocked till auth between Storefront & Checkout is established**


.gitignore
```diff
@@ -48,3 +48,6 @@ cypress/videos
 
 
 **/content/compiled-locales
+
+# saleor-app-sdk - APL for local app development
+.saleor-app-auth.json
```
lkostrowski, line 53: This should be ignored, contain tokens for local development

lkostrowski: TODO - why graphql.schema.json and graphql.index.ts commited, shouldnt they be in gitignore?

apps/storefront/.env
```diff
@@ -1,6 +1,5 @@
-NEXT_PUBLIC_API_URI=$SALEOR_API_URL
+NEXT_PUBLIC_API_URI=https://saleor-demo-pxzg5bvn.saleor.cloud/graphql/
```
lkostrowski, line 1: temp

apps/storefront/components/Navbar/Navbar.tsx
```diff
@@ -24,10 +25,12 @@ export function Navbar() {
...
   const { currentLocale } = useRegions();
+  const { checkoutUrl } = useSaleorAppSettings();
 
-  const externalCheckoutUrl = checkout
-    ? `/checkout/?checkout=${checkout.id}&locale=${currentLocale}`
-    : "#";
+  const externalCheckoutUrl =
+    checkout && checkoutUrl
+      ? `${checkoutUrl}/?checkout=${checkout.id}&locale=${currentLocale}`
+      : "#";
```
lkostrowski, line 33: is this the only place where checkout redirects? Should this logic be extracted somewhere?

apps/storefront/next.config.js
```diff
@@ -49,16 +49,20 @@ module.exports = withBundleAnalyzer({
...
         headers: [{ key: "x-frame-options", value: "ALLOWALL" }],
       },
+      /**
+       * TODO. Ensure this is valid approach. Best if it can be limited to Saleor,
+       *   so maybe it should use NEXT_PUBLIC_API_URI?
+       */
+      {
+        source: "/dashboard/(.*)",
+        headers: [{ key: "content-security-policy", value: "frame-ancestors *" }],
+      },
```
lkostrowski, line 59: This page must be loaded in dashboard, we can lock specific dashboard in header?

apps/storefront/package.json
```diff
@@ -23,13 +23,19 @@
     "@graphql-codegen/typescript-apollo-client-helpers": "2.2.5",
     "@headlessui/react": "1.7.3",
     "@heroicons/react": "1.0.6",
+    "@material-ui/core": "^4.12.4",
+    "@material-ui/icons": "^4.11.3",
+    "@material-ui/lab": "4.0.0-alpha.61",
+    "@saleor/app-sdk": "^0.18.0",
+    "@saleor/macaw-ui": "^0.6.6",
```
lkostrowski, line 30: Macaw is used to build this single page in dashboard, but it requires material to be installed.

Its Next app so I believe there should be no bundle impact on public code, but not sure. Also Macaw/Material can introduce some problems with peers.

Since we only need very simple page in dashboard (1-field form) I can make this page without Macaw. It may not be 100% consistent, but maybe its be better to drop this legacy.

WDYT @mmiszy ?

apps/storefront/package.json
```diff
@@ -47,7 +53,8 @@
     "react-intl": "6.2.1",
     "react-use": "17.4.0",
     "tailwind-scrollbar-hide": "1.1.7",
-    "url-join": "4.0.1"
+    "url-join": "4.0.1",
+    "urql": "^3.0.3"
```
lkostrowski, line 57: At the moment, SDK examples and app-templates use urql. Storefront uses Apollo. How should we approach? We can adapt to Apollo, but maybe there is some migration plan to urql (its sed in checkout?)

Its also backend-only code, so shouldn't affect bundle

apps/storefront/pages/_app.tsx
```diff
@@ -30,10 +31,12 @@ function MyApp({ Component, pageProps }: AppPropsWithLayout) {
       <CheckoutProvider>
         <RegionsProvider>
           <SaleorProviderWithChannels>
-            <BaseSeo />
-            <NextNProgress color="#5B68E4" options={{ showSpinner: false }} />
-            {DEMO_MODE && <DemoBanner />}
-            {getLayout(<Component {...pageProps} />)}
+            <SaleorAppSettingsProvider>
```
lkostrowski, line 34: its wrapping entire App but technically it doesnt have to. I can wrap only required component or even remove Provider at all (if header is only place that needs it, I can run fetch() there)

apps/storefront/pages/api/manifest.ts
```diff
@@ -0,0 +1,17 @@
+import { createManifestHandler } from "@saleor/app-sdk/handlers/next";
+import { AppManifest } from "@saleor/app-sdk/types";
+
+export default createManifestHandler({
+  manifestFactory(context: { appBaseUrl: string }): AppManifest {
+    return {
+      id: "saleor.storefront.app",
+      version: "1.0.0", // is there any official versioning?
```
lkostrowski, line 8: @mmiszy ideas?

apps/storefront/pages/api/settings.ts
```diff
@@ -0,0 +1,89 @@
...
+  // Helper located at `src/lib/metadata.ts` returns manager which will be used to
+  // get or modify metadata.
+  const settings = createSettingsManager(client);
+
+  if (req.method === "GET") {
+    await sendResponse(res, 200, settings);
+    return;
+  } else if (req.method === "POST") {
+    if (!token) {
+      // TODO validate token, probably extract logic from middleware in SDK to standalone function (or make a separate handler)
```
lkostrowski, line 65: @krzysztofwolski FYI, we need to ensure this can be run only from Dashboard, having user token.

apps/storefront/package.json
mmiszy: I'm not sure if there's a migration plan. I don't have any preference for either urql or apollo, so we should just pick one and plan the migration.

apps/storefront/pages/api/settings.ts
```diff
@@ -0,0 +1,89 @@
...
+// Even multiple calls of `get` method will result with only one call to the database.
+const sendResponse = async (
+  res: NextApiResponse<SettingsApiResponse>,
+  statusCode: number,
+  settings: SettingsManager
+) => {
+  res.status(statusCode).json({
+    success: statusCode === 200,
+    data: {
+      checkoutUrl: (await settings.get("checkoutUrl")) || "",
```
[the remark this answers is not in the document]
lkostrowski: I will fix all of these on post-draft :)

apps/storefront/next.config.js
```diff
@@ -49,16 +49,20 @@ module.exports = withBundleAnalyzer({
...
     ];
   },
   async rewrites() {
     const cloudDeploymentUrl = process.env.CLOUD_DEPLOYMENT_URL;
 
     return [
-      {
-        source: "/checkout/",
-        destination: `${process.env.NEXT_PUBLIC_CHECKOUT_URL}/`,
```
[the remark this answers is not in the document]
lkostrowski: Can you explain why? I looked on the codebase and I replaced a single instance of this redirect here:
https://github.com/saleor/react-storefront/pull/592/files#diff-e69091281753d50627a3bc1999dbb8e8abde136c55a1d841a299431142d00077L29

(at least I haven't found more)

This operation (fetching redirect URL) can be cached and is almost one-time, when user wants to redirect to the checkout. I don't see any performance issues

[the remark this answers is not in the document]
lkostrowski: I see. Middlewares sound good. Still I think it will only need to fetch this URL once, for this single call, right?

We can even fetch redirect URL on frontend and then redirect to `/checkout?checkoutURL=${fetchedCheckoutUrl}&checkoutId=${someID}`, which will be intercepted by middleware and rewritten?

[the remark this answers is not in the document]
mmiszy: Each request to `/checkout` will trigger the middleware because each od them needs to be rewritten.
You can try but I think every script, static file, all medias, and api calls will go through it and it'll be slow.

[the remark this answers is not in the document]
lkostrowski: What are other options?

apps/storefront/pages/api/settings.ts
krzysztofwolski: Issue created https://github.com/saleor/saleor-app-sdk/issues/104

apps/storefront/pages/api/manifest.ts
lkostrowski: I think we should introduce some incremental versioning one day. Doesnt have to be semver, but I assume we will end up with:

1. Apps with breaking changes, eg. "Checkout v4 works with Saleor >=4.x"
2. Clear versioning for semi-technical people, where its easy to compare that version 3.1 is newer than 3.0

Commits are good as build ids for Sentry but not so good for end users

.gitignore
mmiszy: I don't think we use graphql.schema.json so it can be removed completely. graphql.index.ts should stay.

apps/storefront/next.config.js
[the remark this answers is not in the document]
mmiszy: Rethinking our auth flow for cross domain requests.

lkostrowski:
Closing at the moment, we are not ready to implement this, and when we are, PR will be stale anyway

PR #601, opened 2022-11-03
Add phone input

### LINKED ISSUE #371: Add phone input to checkout
- [x] Add phone number input to the Checkout flow.
- [x] Validate phone number.

In checkout designs there is a [nice phone input](https://www.figma.com/file/oBQhZFwyFcXauxFrb3mbRb/saleor_checkout?node-id=2482%3A11686). Let's implement it in ui-kit so it's more user friendly.

bmigirl:
Fixes #371


packages/checkout-storefront/package.json
```diff
@@ -44,7 +44,9 @@
     "checkout-common": "workspace:*",
     "clsx": "1.2.1",
     "concurrently": "7.5.0",
+    "fs": "0.0.1-security",
```
mmiszy, line 47: What does it do?

packages/checkout-storefront/src/components/AddressForm/useAddressFormValidationResolver.ts
```diff
@@ -0,0 +1,67 @@
...
+        });
+
+        result = {
+          ...result,
+          values,
+        };
+      } catch (error) {
+        const errors = getErrorsAsObject(getAllValidationErrors(error as YupValidationErrorObject));
+        result = { ...result, errors };
+      }
```
mmiszy, line 46: I think we could extract it to a function `getAddressFormResolverValue` or something, and avoid using `let` then.

packages/checkout-storefront/src/components/AddressForm/useAddressFormValidationResolver.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      const phoneRequiredError = result.errors?.phone?.type === "required";
+
+      if (!isValidNumber(data.phone, data.countryCode) && !phoneRequiredError) {
+        result = {
```
mmiszy, line 51:
```suggestion
        return {
```

packages/checkout-storefront/src/hooks/__tests__/usePhoneNumberUtils.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  it("should return false for invalid phone numbers", () => {
+    const { result: hook } = renderHook(() => usePhoneNumberUtils());
+
+    expect(hook.current.isValidNumber("+4172268096", countryCode)).toEqual(false);
+    expect(hook.current.isValidNumber("+48111111111", countryCode)).toEqual(false);
+    expect(hook.current.isValidNumber("123", countryCode)).toEqual(false);
+    expect(hook.current.isValidNumber("foobar", countryCode)).toEqual(false);
+    expect(hook.current.isValidNumber("+48", countryCode)).toEqual(false);
+  });
+});
```
mmiszy, line 23: Let's also add tests for `getPhoneNumberWithCountryCode`

packages/checkout-storefront/src/hooks/usePhoneNumberUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+        return `+${callingCode}${phone}`;
+      }
+
+      return phone;
+    },
+    [getPhoneNumberInstance, isMissingCountryCallingCode]
+  );
+
+  return { isValidNumber, getPhoneNumberWithCountryCode };
+};
```
mmiszy, line 57: I think this doesn't even need to be a React hook. Just plain old utility functions. What do you think? No need for `useCallback`s and easier to test.

mmiszy (changes requested):

packages/checkout-storefront/src/components/AddressForm/useAddressFormValidationResolver.ts
bmigirl: tbh I added it like that knowing that it's like the resolvers will be refactored soon along with changing the forms lib so for now I'd just leave it as it is

packages/checkout-storefront/src/hooks/usePhoneNumberUtils.ts
bmigirl: Oh yeah, you're right. I added it like this when testing with another phone number lib and it required to be hook, later forgot to change it.

mmiszy (dismissed):

mmiszy:
You forgot to run `pnpm i` after removing that one dependency, other than that looks good!

packages/checkout-storefront/src/lib/utils/phoneNumber.ts
```diff
@@ -0,0 +1,39 @@
...
+  } catch (error) {
+    return null;
+  }
+};
+
+export const isValidPhoneNumber = (phone: string, countryCode: CountryCode | undefined) =>
+  !!getPhoneNumberInstance(phone, countryCode)?.isValid();
+
+const isMissingCountryCallingCode = (phone: string, countryCode: CountryCode) => {
+  const isValidWithoutCountryCode = getPhoneNumberInstance(phone, undefined);
```
witoszekdev, line 25: Why do we use `getPhoneNumberInstance` instead of `isValidPhoneNumber` here?

michalina-graczyk:
When I go to checkout there is 404. Please check https://watch.screencastify.com/v/bZottsmZZMUV2EWhCAA0

mmiszy (approved):

PR #629, opened 2022-11-18
#286 Forbid imports across packages and apps

### LINKED ISSUE #286: Configure `eslint` to prevent imports between apps
We can configure `eslint` to prevent accidental imports from one app to another.
This will help us keep our codebase clean, builds fast and code reusable.

Common types and utilities should be moved to [packages/checkout-common](https://github.com/saleor/saleor-checkout/tree/main/packages/checkout-common) instead.

mmiszy:
Fixes #286

### DIFF
--- a/packages/checkout-common/src/utils.ts
+++ b/packages/checkout-common/src/utils.ts
@@ -1,4 +1,10 @@
+export const anythingToString = <Anything>(value: Anything) => {
+  try {
+    return JSON.stringify(value);
+  } catch (err) {
+    return String(value);
+  }
+};
 export function assertUnreachable(value: never): never {
-  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-  throw new Error(`Unexpected case: ${value}`);
+  throw new Error(`Unexpected case: ${anythingToString(value)}`);
 }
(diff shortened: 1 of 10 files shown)


packages/checkout-common/src/utils.ts
```diff
@@ -1,3 +1,4 @@
 export function assertUnreachable(value: never): never {
+  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
   throw new Error(`Unexpected case: ${value}`);
```
witoszekdev, line 3: Why can't we do this?
```suggestion
  throw new Error(`Unexpected case: ${value as string}`);
```

witoszekdev (commented):
nice 👍🏻

packages/checkout-common/src/utils.ts
mmiszy: We can but why is that better? `value` is not a string, it could be anything.

witoszekdev: You're right, that just avoids the problem, maybe we should provide a default if `value` is not a string - I think this ESLint rule makes sense, as we don't want to have `object [Object]` in an error 😅

witoszekdev (approved):

PR #641, opened 2022-11-21
#589 Fix Hydration error

### LINKED ISSUE #589: Hydration error
### Summary

Hydration error is visible when `saleor-app-checkout` is loaded in iframe in development

CleanShot 2022-10-26 at 17 27 51@2x

### Steps to reproduce

1. Start local Saleor instance using `saleor-platform`: `docker-compose up api`
2. Install app in Saleor which uses tunnel
3. Visit app configuration page

_No response_

### Output from envinfo

_No response_

mmiszy:
Render `saleor-app-checkout` on the client.
It's more or less the same thing that [saleor-app-template does](https://github.com/saleor/saleor-app-template/blob/f05dbcf9a3a0bfe5081e4181693655f22a8e3f9e/src/pages/index.tsx#L7).

Fixes #589


apps/saleor-app-checkout/frontend/hooks/useGetPaymentProviderSettings.ts
```diff
@@ -12,18 +12,19 @@ export const useGetPaymentProviderSettings = <TArgs>(
...
 
-  const domain = app?.getState().domain;
-  if (!domain) {
-    console.error(`Missing domain!`);
-  }
-  const saleorApiUrl = domain ? `https://${domain}/graphql/` : "";
+  const { app } = useAppContext();
+  const domain = app.getState().domain;
+
+  // @todo use saleorApiUrl
```
bmigirl, line 18: is this still todo or done?

apps/saleor-app-checkout/frontend/hooks/useSetPaymentProviderSettings.ts
```diff
@@ -10,18 +10,18 @@ export const useSetPaymentProviderSettings = <TArgs>(
...
   const { privateSettings, setPrivateSettings } = usePrivateSettings();
 
-  const domain = app?.getState().domain;
-  if (!domain) {
-    console.error(`Missing domain!`);
-  }
-  const saleorApiUrl = domain ? `https://${domain}/graphql/` : "";
+  const { app } = useAppContext();
+  const domain = app.getState().domain;
+  // @todo use saleorApiUrl
```
bmigirl, line 15: same

apps/saleor-app-checkout/frontend/components/elements/AppProvider/ClientAppBridgeProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+          to: `/apps/${encodeURIComponent(app.getState().id)}/app${rawPath}`,
+        })
+      );
+    };
+
+    const unsubscribe = app.subscribe("redirect", ({ path }) => {
+      if (path && path !== router.pathname) {
+        void router.push(path);
+      }
+    });
```
witoszekdev, line 59: This seems a bit odd: we're subscribing to a redirect to... unsubscribe? Does that mean we're registering a new event handler with app-sdk?
If that so, then that's a potential memory leak, since this component will be unmounted (`unsubscribe` is run in a returned function from `useEffect`)

apps/saleor-app-checkout/frontend/fetch/requests.ts
```diff
@@ -1,35 +1,42 @@
...
+  token: string;
 }): FetchResponse<PaymentProviderSettingsResult> =>
   fetch(
     `/api/payment-provider-settings/` + `?` + new URLSearchParams({ saleorApiUrl }).toString(),
     {
       method: "GET",
-      headers: getAuthHeaders(),
+      headers: {
+        Authorization: token ? `Bearer ${token}` : "",
+      },
```
witoszekdev, line 21: Why did we remove this function?

apps/saleor-app-checkout/frontend/hooks/useGetPaymentProviderSettings.ts
mmiszy: Still todo:
https://github.com/saleor/saleor-app-sdk/issues/87
https://github.com/saleor/saleor/issues/10819

apps/saleor-app-checkout/frontend/hooks/useSetPaymentProviderSettings.ts
mmiszy: Also todo:
https://github.com/saleor/saleor-app-sdk/issues/87
https://github.com/saleor/saleor/issues/10819

apps/saleor-app-checkout/frontend/fetch/requests.ts
mmiszy: Because it didn't really make any sense. It returned `{ Authorization: "Bearer ………" }` on the client and `{Authorization: ""}` on the server. It's confusing and against best practices.

apps/saleor-app-checkout/frontend/components/elements/AppProvider/ClientAppBridgeProvider.tsx
mmiszy: No, I think it's fine. The component mounts and subscribes to the `redirect` event from appBridge.
When the component unmounts, it unsubscribes.

mmiszy: Refactored this a bit

witoszekdev (approved):

matteius:
When I pulled in these changes it broke my checkout app because it no longer pulls the app token from the `.saleor-app-auth.json`  -- I spent some time verifying this.   Related question -- is the checkout app token supposed to be a secret value or is it publicly exposed in the app?

mmiszy:
@matteius what do you mean it doesn't pull the token from `.saleor-app-auth.json`? Can you create a bug report? Thanks.

The token is a secret.

matteius:
@mmiszy  Sure, I'll make a report--the gist is I had added a `.saleor-app-auth.json` file to define the app token in `apps/saleor-app-checkout/`.saleor-app-auth.json` which was working, but I found when I pulled in this change that it no longer got the app token (it was undefined at the point of use).   Also, given the token is a secret, wouldn't changing to client side rendering possibly expose what that value is to the end-user?

mmiszy:
@matteius thanks for taking the time to investigate this! I'm glad that it worked in the end.

> given the token is a secret, wouldn't changing to client side rendering possibly expose what that value is to the end-user?

Token is a secret so we never use the token on the client-side. It's only used in API routes on the backend.

PR #646, opened 2022-11-23
Add zustand 

### LINKED ISSUE #640: Add Zustand

bmigirl:
Fixes #640


packages/checkout-storefront/src/components/AutoSaveAddressForm/AutoSaveAddressForm.tsx
```diff
@@ -4,23 +4,34 @@ import {
...
+
   const { formProps, onSubmit: handleSubmit } = useAddressForm({
     defaultValues,
     onSubmit,
   });
   const { getValues } = formProps;
 
+  useCheckoutFormValidationTrigger({
+    formProps,
+    scope: isShippingForm ? "billingAddress" : "shippingAddress",
```
mmiszy, line 32: Is this correct? looks like it should be the other way round 🤔

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutFinalize.ts
```diff
@@ -1,20 +1,17 @@
...
+} from "@/checkout-storefront/state/paymentDataStore";
 
 export const useCheckoutFinalize = () => {
   const { checkout } = useCheckout();
-  const { register } = useAuth();
-  const { user } = useAuthState();
   const { checkoutPay, loading, error: payError, data: _payData } = usePay();
-  const { showErrors, showCustomErrors } = useAlerts();
-  const { errors, setApiErrors } = useErrors<CheckoutFormData>();
+  // const { showCustomErrors } = useAlerts();
```
mmiszy, line 13:
```suggestion
```

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutFinalize.ts
```diff
@@ -1,20 +1,17 @@
...
+  const { paymentMethod, paymentProvider } = useSelectedPaymentData() as SelectedPaymentData;
```
mmiszy, line 14: This assertion is not correct. The values could be `null`.

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutSubmit.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const updateStateValues = Object.values(updateState);
+
+  const anyRequestsInProgress =
+    updateStateValues.some((status) => status === "loading") || loadingCheckout;
+
+  const finishedApiChangesWithNoError =
+    !anyRequestsInProgress && updateStateValues.every((status) => status === "success");
+
+  const allFormsValid =
```
mmiszy, line 35: Maybe we should have hooks for these utilities?

```
useAreAnyRequestsInProgres()
useFinishedApiChangesWithNoError()
useAreAllFormsValid()
```

etc.

packages/checkout-storefront/src/sections/PaymentSection/PaymentMethods.tsx
```diff
@@ -91,7 +23,9 @@ export const PaymentMethods: React.FC<PaymentMethodsProps> = ({
           className="shrink"
           value={paymentMethodId}
           selectedValue={selectedPaymentMethod || availablePaymentMethods[0]}
-          onChange={onSelect}
+          onChange={(event: ChangeEvent<HTMLInputElement>) =>
```
mmiszy, line 26:
```suggestion
          onChange={(event) =>
```

packages/checkout-storefront/src/state/updateStateStore.ts
```diff
@@ -0,0 +1,107 @@
...
+      loadingCheckout,
+    }),
+    shallow
+  );
+
+  return useMemo(
+    () => ({ updateState, loadingCheckout }),
+    // because we want to compare array of strings instead of object
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [loadingCheckout, ...Object.values(updateState)]
```
mmiszy, line 82: Why do we want to compare an array of strings? Won't it depend on the order in which values are updated in the store, i.e.
`{ checkoutShippingUpdate: "success", checkoutCustomerAttach: "error" }`
`{ checkoutCustomerAttach: "error", checkoutShippingUpdate: "success" }`
will be treated as not equal?

mmiszy, line 82: Why do we compare Object.values(…) here? Won't this depend on the order in which fields are updated in the store, i.e.

`{ checkoutAddPromoCode: "error", userRegister: "success" }`
and
`{ userRegister: "error", checkoutAddPromoCode: "success" }`
will be treated as equal?

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutSubmit.ts
```diff
@@ -0,0 +1,61 @@
...
+import { useCheckoutFinalize } from "@/checkout-storefront/sections/CheckoutForm/useCheckoutFinalize";
+
+export const useCheckoutSubmit = () => {
+  const { validateAllForms } = useCheckoutValidationActions();
+  const { validating, validationState } = useCheckoutValidationState();
+  const { updateState, loadingCheckout } = useCheckoutUpdateState();
+  const { setShouldRegisterUser } = useCheckoutUpdateStateActions();
+  const { checkoutFinalize, finalizing } = useCheckoutFinalize();
+
+  const [submitInProgress, setSubmitInProgress] = useState(false);
```
mmiszy, line 19: Wondering if it makes sense to move this value to the store as well, just to be consistent.

Then we could make a hook like `useIsCheckoutProcessing`

mmiszy:
I think this is the right direction.

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutSubmit.ts
bmigirl: I don't think that there is a good reason to extract those to hooks. They're basically one line conditions so I'd say a hook for each would be an overkill. If it gets more crowded here I'll extract them to one utility hook tho

packages/checkout-storefront/src/sections/PaymentSection/PaymentMethods.tsx
bmigirl: We need it here, otherwise ts screms for no "value" inside this handler

packages/checkout-storefront/src/sections/CheckoutForm/useCheckoutSubmit.ts
bmigirl: I'd say no since it's only used in this one place and passed down to display the button loader so it seems to me more like a component state related than checkout itself

packages/checkout-storefront/src/components/AutoSaveAddressForm/AutoSaveAddressForm.tsx
bmigirl: It's not, changing

packages/checkout-storefront/src/state/updateStateStore.ts
bmigirl: Actually, I don't think we need to useMemo here at all. It seems like it doesn't do much in this case

packages/checkout-storefront/src/sections/Contact/useGuestUserForm.ts
```diff
@@ -0,0 +1,165 @@
...
+import {
+  extractMutationErrors,
+  localeToLanguageCode,
+  useValidationResolver,
+} from "@/checkout-storefront/lib/utils";
+import { useAuthState } from "@saleor/sdk";
+import { useCallback, useEffect, useState } from "react";
+import { useForm } from "react-hook-form";
+import { object, string } from "yup";
+// import { errorMessages as passwordErrorMessages } from "@/checkout-storefront/hooks/useErrorMessages/messages";
```
mmiszy, line 22:
```suggestion
```

packages/checkout-storefront/src/sections/Contact/useGuestUserForm.ts
```diff
@@ -0,0 +1,165 @@
...
+
+export interface GuestUserFormData {
+  email: string;
+  password: string;
+}
+
+export const useGuestUserForm = ({ createAccount }: { createAccount: boolean }) => {
+  // const formatMessage = useFormattedMessages();
```
mmiszy, line 30:
```suggestion
```

mmiszy (changes requested):

mmiszy (approved):

PR #666, opened 2022-12-07
REST APL and generic register handler

krzysztofwolski:
- add new SALEOR CLOUD APL
- use [generic register handler](https://github.com/saleor/saleor-app-sdk/blob/main/docs/api-handlers.md#app-register-handler-factory) 
- use request validation for restricted access handlers (and removed duplicated JWT verification code)
- drop deprecated VercelAPL


lkostrowski (approved):
LGTM 😈

krzysztofwolski:
image

#666 FINALLY

witoszekdev (approved):

lkostrowski:
Hey @saleor/storefront-checkout please take a look - we should merge this, because its already a long living branch and we will have problem to maintain it

mmiszy:
I'll review it today @lkostrowski

apps/saleor-app-checkout/config/apl.ts
```diff
@@ -1,62 +1,60 @@
...
-        return new UpstashAPL({
-          restURL: UPSTASH_REDIS_REST_URL,
-          restToken: UPSTASH_REDIS_REST_TOKEN,
-        });
-      })();
-    case "file":
-      void printFileAplWarning();
-      return new FileAPL();
-    case "vercel":
-      return new CheckoutVercelAPL();
```
mmiszy, line 31: We're currently using this on staging and demo. What should we do now?

apps/saleor-app-checkout/config/apl.ts
```diff
@@ -1,62 +1,60 @@
...
 import invariant from "ts-invariant";
 import Fs from "fs/promises";
 import { unpackPromise } from "../utils/unpackErrors";
-import { CheckoutVercelAPL } from "./checkoutVercelApl";
 
-type Result = {
-  saleorApiUrl: string;
-  appToken: string;
-};
+export let apl: APL;
```
mmiszy, line 6:
```suggestion
export const apl = getApl();
```

apps/saleor-app-checkout/pages/api/set-payment-provider-settings.ts
```diff
@@ -52,4 +43,5 @@ const handler: NextApiHandler = async (req, res) => {
     return res.status(500).json({ error });
   }
 };
-export default allowCors(requireAuthorization(handler, ["HANDLE_PAYMENTS"]));
```
mmiszy, line 55: Where do we specify that `HANDLE_PAYMENTS` permission is required for accessing this route?

mmiszy (changes requested):

apps/saleor-app-checkout/config/apl.ts
krzysztofwolski: Demo instance should be using SaleorCloud APL.
For custom deployments Upstash APL can be used.

In the future, a workflow will cover reading tokens from the environment. Before we'll be able to implement it, more changes in the core are necessary.

krzysztofwolski:
@mmiszy 
- added requested changes
- resolved conflicts 
- the build step fails because of the misconfiguration of APL. Could you set up the keys I generated last week?

krzysztofwolski:
Pushed update with the newest version of the SDK. No code changes were needed.

apps/saleor-app-checkout/config/apl.ts
lkostrowski: There is a way to do it, but not "dynamically" with VercelAPL. This approach is a rabbit hole and we dropped it some time ago
We will work on the documentation. Officially we encourage to use other APL.

To achieve it for single tenant way (which we don't recommend):
1. console.log authdata in registerHandler
2. save it to env variables
3. write simple read-only "EnvApl" and read it from there

We don't recommend it because this flow will break with upcoming features like token rotation or upgrades

Hence even single tenant flow should use some kind of mutable storage

mmiszy: But that's exactly what `CheckoutVercelAPL` did – read variables from env. So why is it removed?

lkostrowski: I just checked the code, indeed, it may be a mistake on our side
@krzysztofwolski can you check it again

Broader context - we removed build-in `VercelAPL` from SDK, which was actually complex and fragile logic, that was using external service (appraptor) hosted by us, that requires Vercel tokens to talk to Vercel API and update env variables on the fly. It was hacky and often broke.

In the future we want to enable Web API + VercelEdgeConfig to re-enable single tenant flow with Vercel

lkostrowski: Just talked to @krzysztofwolski  - we need to find a middle spot. Approach you did have in CheckoutVercelAPL is fine, but it needs more fields - Krzyh will update it

krzysztofwolski: @mmiszy Checkout Vercel APL is back. I've added fields required by the new APL interface.

krzysztofwolski:
@mmiszy
- rebased with the canary branch
- CheckoutVercelAPL is back and updated
- found a bunch of no longer-used code and dropped it

PR #701, opened 2023-01-17
Add Slovak translation

andrewshadura:
This translation is currently missing some strings, I’ll push an update a bit later.


andrewshadura:
It’s June now, any updates? Thanks!

andrewshadura:
@mmiszy, an explanation would be great?

PR #734, opened 2023-02-08
Rewrite auth

### LINKED ISSUE #714: Refactor auth in checkout
```[tasklist]
### Tasks
- [x] Remove `@saleor/sdk`
- [x] Implement authentication flow https://docs.saleor.io/docs/3.x/api-usage/authentication
```

> Since Saleor SDK is no longer maintained we need to remove it altogether from checkout. We already have following stuff replaced: passwordResetRequest, passwordReset and userRegister. We need to implement the login flow because currently we're using: login, logout and authState with user and authenticating props. That'll also allow us to remove anything connected to apollo client.

bmigirl:
Fixes #714 
[Rewrite auth in checkout](https://github.com/saleor/react-storefront/issues/714)


packages/checkout-storefront/src/lib/auth/useAuthChange.ts
```diff
@@ -0,0 +1,36 @@
...
+      onSignedIn();
+    } else if (authState === "signedOut") {
+      onSignedOut();
+    }
+  };
+
+  useEffect(() => {
+    // for current window
+    window.addEventListener(STORAGE_AUTH_EVENT_KEY, handleAuthChange as EventListener);
+    // //  for other windows
```
mmiszy, line 31:
```suggestion
    window.addEventListener(STORAGE_AUTH_EVENT_KEY, handleAuthChange as EventListener);
```

packages/checkout-storefront/src/lib/auth/useSaleorAuthClient.ts
```diff
@@ -0,0 +1,39 @@
...
+  const saleorAuthClient = useMemo(
+    () =>
+      new SaleorAuthClient({
+        storage,
+        saleorApiUrl,
+        onAuthRefresh: (value) => {
+          setIsAuthenticating(value);
+          if (typeof onAuthRefresh === "function") {
+            onAuthRefresh(value);
+          }
```
mmiszy, line 25:
```suggestion
          onAuthRefresh?.(value);
```

packages/checkout-storefront/src/lib/auth/useSaleorAuthClient.ts
```diff
@@ -0,0 +1,39 @@
+import { SaleorAuthClient, SaleorAuthClientProps } from "./SaleorAuthClient";
+import { useEffect, useMemo, useState } from "react";
+
+export interface UseSaleorAuthClient {
+  saleorAuthClient: SaleorAuthClient;
+  isAuthenticating: boolean;
+}
+
+export const useSaleorAuthClient = ({
```
mmiszy, line 9: I think there could be a temptation to call this hook in multiple components – and that won't work 😱 
I think we can try making the intention here clearer by renaming it to something like:
```suggestion
/**
 * This hook should be used only once per application.
 * @see Use {@link useSaleorAuthContext} for accessing the auth state.
 */
export const useCreateNewSaleorAuthClient = ({
```

packages/checkout-storefront/src/hooks/useUser.ts
```diff
@@ -0,0 +1,12 @@
+import { useUserQuery } from "@/checkout-storefront/graphql";
+import { useMemo } from "react";
+
+export const useUser = () => {
+  const [{ data, fetching: loading, stale }] = useUserQuery();
+
+  const user = data?.user;
+
+  const authenticated = useMemo(() => !!user?.id, [user?.id]);
```
mmiszy, line 9: This doesn't need `useMemo`:
```suggestion
  const authenticated = !!user?.id;
```

mmiszy (changes requested):

packages/checkout-storefront/src/lib/auth/useSaleorAuthClient.ts
bmigirl: Thought making it clear in the docs would be sufficient but sure 👍

witoszekdev: We could also add a check if a client instance was created inside the project:

```tsx
let clientCreated = true;

export const useSaleorAuthClient = () => {
  // ...
}

```

witoszekdev (commented):
Great work 😄 
I really liked the docs ;)

packages/checkout-storefront/src/lib/auth/SaleorAuthClient.ts
```diff
@@ -0,0 +1,190 @@
...
+      return readResponse;
+    }
+
+    if (token) {
+      this.accessToken = token;
+    }
+
+    if (refreshToken) {
+      this.storageHandler.setRefreshToken(refreshToken);
+    }
```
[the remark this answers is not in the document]
bmigirl: I mean technically there shouldn't be such a case. It should either be `errors: empty array + token: string + refreshToken: string` or `errors: filledArray or token: null / undefined refreshToken: null / undefined`, so I guess I would assume here that api works correct?

packages/checkout-storefront/src/lib/auth/SaleorAuthClient.ts
```diff
@@ -0,0 +1,190 @@
...
+  };
+
+  private handleRequestWithTokenRefresh: Fetch = async (input, init) => {
+    const refreshToken = this.storageHandler.getRefreshToken();
+
+    invariant(refreshToken, "Missing refresh token in token refresh handler");
+
+    // the refresh already, finished, proceed as normal
+    if (this.accessToken) {
+      this.tokenRefreshPromise = null;
```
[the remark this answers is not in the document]
bmigirl: Nah - had we removed it right away after the request is done and before the access token is set, there'd be a moment when another call could run that function, causing another refresh request because there'd be no access token and no refresh promise present at that moment. But I think we could move it to line 101, like so:
```
  this.accessToken = token;
  this.tokenRefreshPromise = null;
  return this.runAuthorizedRequest(input, init);
```

packages/checkout-storefront/src/lib/auth/SaleorAuthClient.ts
```diff
@@ -0,0 +1,190 @@
...
+
+export class SaleorAuthClient {
+  private accessToken: string | null = null;
+  private tokenRefreshPromise: null | Promise<Response> = null;
+  private onAuthRefresh?: (isAuthenticating: boolean) => void;
+  private saleorApiUrl: string;
+  private storageHandler: SaleorAuthStorageHandler;
+  cleanup: SaleorAuthStorageHandler["cleanup"];
+
+  constructor({ saleorApiUrl, storage, onAuthRefresh }: SaleorAuthClientProps) {
```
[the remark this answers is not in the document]
mmiszy: Storage is already an interface, not sure what do you mean?
Accepting AsyncStorage would require more changes than just the type.

packages/checkout-storefront/src/lib/auth/SaleorAuthClient.ts
```diff
@@ -0,0 +1,190 @@
...
+    this.storageHandler = new SaleorAuthStorageHandler(storage);
+    this.onAuthRefresh = onAuthRefresh;
+    this.saleorApiUrl = saleorApiUrl;
+    this.cleanup = this.storageHandler.cleanup;
+  }
+
+  private runAuthorizedRequest: Fetch = (input, init) => {
```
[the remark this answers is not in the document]
mmiszy: Why? `private` makes it readable enough.

packages/checkout-storefront/src/lib/auth/useSaleorAuthClient.ts
mmiszy: That wouldn't work correctly in many cases.

witoszekdev (dismissed):

mmiszy (dismissed):

michalina-graczyk:
Steps:
1. Go to checkout 
2. Click log in 
3. Click on forgot password 
Results on the screen:

image

michalina-graczyk:
Password requirements are not defined 
1. Create new account
2. Click on the confirmation link from your mail 
3. Type new pass eg. test888
4. Fill all billing address details
5. Click 'pay' button
Current result: After filling all required fields I am getting a message that my password is too short, it would be better to have this requirments at the beggining of creating pass.

bmigirl:
Because we need to get on with the auth and these ui fixes are not exactly related with this task, (also we're awaiting decision from design on how these fixes should look), created a [ticket](https://github.com/saleor/react-storefront/issues/754) and merging

mmiszy (approved):

PR #755, opened 2023-02-21
Rewrite auth in storefront

### LINKED ISSUE #715: Refactor auth in storefront
```[tasklist]
### Tasks
- [ ] Remove `@saleor/sdk`
- [ ] Implement authentication flow https://docs.saleor.io/docs/3.x/api-usage/authentication
```

bmigirl:
Fixes #715


apps/storefront/pages/[channel]/[locale]/index.tsx
```diff
@@ -5,22 +5,23 @@ import React, { ReactElement } from "react";
...
 export const getStaticProps = async (context: GetStaticPropsContext) => {
-  const result: ApolloQueryResult<HomepageBlocksQuery> = await apolloClient.query<
+  const result: ApolloQueryResult<HomepageBlocksQuery> = await staticApolloClient.query<
     HomepageBlocksQuery,
     HomepageBlocksQueryVariables
   >({
     query: HomepageBlocksQueryDocument,
     variables: { slug: HOMEPAGE_MENU, ...contextToRegionQuery(context) },
   });
+  console.log({ result });
```
Comment (witoszekdev) [inline apps/storefront/pages/[channel]/[locale]/index.tsx:24]: Console.log left

apps/storefront/lib/auth/useApolloClient.ts
```diff
@@ -0,0 +1,31 @@
+import { Fetch } from "@/lib/auth/types";
+import { API_URI } from "@/lib/const";
+import { ApolloClient, createHttpLink, InMemoryCache } from "@apollo/client";
+import { useMemo } from "react";
+
+import { typePolicies } from "./typePolicies";
+
+// for static geenration of pages, we don't need auth there
+export const staticApolloClient = new ApolloClient({
```
witoszekdev, line 9: Why is it called `static`? Since its purpose is to make unauthenticated requests from backend, maybe we should call it `severApolloClient` or `unauthenticatedAppolloClient`?

apps/storefront/lib/auth/useAuthChange.ts
```diff
@@ -0,0 +1,40 @@
...
+import { SaleorAuthEvent, STORAGE_AUTH_EVENT_KEY } from "./SaleorAuthStorageHandler";
+
+interface UseAuthChangeProps {
+  onSignedIn?: () => void;
+  onSignedOut?: () => void;
+}
+
+// used to handle client cache invalidation on login / logout and when
+// token refreshin fails
+export const useAuthChange = ({ onSignedOut, onSignedIn }: UseAuthChangeProps) => {
```
witoszekdev, line 11: Isn't this the same logic we've added for checkout?
Maybe it should be extracted into a separate package in monorepo?

apps/storefront/lib/auth/useSaleorAuthClient.ts
```diff
@@ -0,0 +1,42 @@
...
+  saleorAuthClient: SaleorAuthClient;
+  isAuthenticating: boolean;
+}
+
+/**
+ * This hook should be used only once per application.
+ * @see Use {@link useSaleorAuthContext} for accessing the auth state.
+ */
+
+export const useSaleorAuthClient = ({
```
witoszekdev, line 14: Isn't this the same logic we've added for checkout?
Maybe it should be extracted into a separate package in monorepo?

apps/storefront/lib/auth/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  query: ReturnType<typeof gql>,
+  variables: TVars
+) => ({
+  method: "POST",
+  headers: {
+    "Content-Type": "application/json",
+  },
+
+  body: JSON.stringify({ query: print(query as unknown as DocumentNode), variables }),
+});
```
witoszekdev, line 37: Isn't this the same logic we've added for checkout?
Maybe it should be extracted into a separate package in monorepo?

apps/storefront/lib/auth/SaleorAuthStorageHandler.ts
```diff
@@ -0,0 +1,60 @@
...
+export const STORAGE_AUTH_EVENT_KEY = "saleor_storage_auth_change";
+export const STORAGE_AUTH_STATE_KEY = "saleor_auth_module_auth_state";
+
+const REFRESH_TOKEN_KEY = "saleor_auth_module_refresh_token";
+
+export type AuthState = "signedIn" | "signedOut";
+
+export type SaleorAuthEvent = CustomEvent<{ authState: AuthState }>;
+
+export class SaleorAuthStorageHandler {
```
witoszekdev, line 11: Isn't this the same logic we've added for checkout?
Maybe it should be extracted into a separate package in monorepo?

apps/storefront/lib/auth/useApolloClient.ts
bmigirl: Indeed, serverApolloClient sounds better. Didn't really figure out a great name here so sure 😅

apps/storefront/lib/auth/useAuthChange.ts
bmigirl: It's literally copy paste, @mmiszy proposed we leave it like that for now. I guess it makes sense since the devtools team might soon want to make a whole separate project out of this.

packages/checkout-storefront/src/lib/auth/SaleorAuthClient.ts
```diff
@@ -67,7 +63,7 @@ export class SaleorAuthClient {
   };
 
   private handleRequestWithTokenRefresh: Fetch = async (input, init) => {
-    const refreshToken = this.storageHandler.getRefreshToken();
+    const refreshToken = this.storageHandler?.getRefreshToken();
 
     invariant(refreshToken, "Missing refresh token in token refresh handler");
```
mmiszy, line 68: We allow `storageHandler` to be `undefined` but then this line will inevitably fail. We should either fix it or change it back to make `storage` required.

mmiszy (changes requested):

packages/checkout-storefront/src/lib/auth/SaleorAuthClient.ts
bmigirl: Technically it shouldn't if this class flow works - this function runs only if you already have refreshToken, check `fetchWithAuth` method, but because refreshToken is not passed as a prop, but pulled from the class instance properties, according to typescript it could be null. So in case you're running this function without a refresh token it should fail because that means the whole flow is broken and it needs to be fixed

witoszekdev (approved):

mmiszy (approved):

michalina-graczyk:
QA approved I didn't find any issues

PR #770, opened 2023-03-07
Various small fixes

bmigirl:
Fixes: 
- saleor auth client falling in loop when refresh token is expired
- "billing address not set" error being thrown within payment in some cases
- summary closing unexpectedly after opening on mobile view
- remove some leftover console.logs

Also added a skeleton for delivery methods. Currently when logged in user has no shipping address set but the auto selection is in progress, the section says "please fill in the shipping address first" - while that makes sense for the guest user, for signed in one it's already happening so it makes more sense to replace the bit with skeleton that'll actually indicate loading to user instead of requirement for action.

### DIFF
--- a/packages/checkout-storefront/src/lib/auth/SaleorAuthClient.ts
+++ b/packages/checkout-storefront/src/lib/auth/SaleorAuthClient.ts
@@ -81,6 +81,7 @@ export class SaleorAuthClient {
       const res: TokenRefreshResponse = await response.clone().json();
 
       const {
+        errors: graphqlErrors,
         data: {
           tokenRefresh: { errors, token },
         },
@@ -88,8 +89,9 @@ export class SaleorAuthClient {
 
       this.onAuthRefresh?.(false);
 
-      if (errors.length || !token) {
-        this.storageHandler?.setAuthState("signedOut");
+      if (errors.length || graphqlErrors?.length || !token) {
+        this.tokenRefreshPromise = null;
+        this.storageHandler?.clearAuthStorage();
         return fetch(input, init);
       }
 
--- a/packages/checkout-storefront/src/sections/DeliveryMethods/DeliveryMethods.tsx
+++ b/packages/checkout-storefront/src/sections/DeliveryMethods/DeliveryMethods.tsx
@@ -11,12 +11,17 @@ import { CommonSectionProps } from "@/checkout-storefront/lib/globalTypes";
 import { deliveryMethodsLabels, deliveryMethodsMessages } from "./messages";
 import { useDeliveryMethodsForm } from "@/checkout-storefront/sections/DeliveryMethods/useDeliveryMethodsForm";
 import { FormProvider } from "@/checkout-storefront/providers/FormProvider";
+import { useCheckoutUpdateState } from "@/checkout-storefront/state/updateStateStore";
+import { DeliveryMethodsSkeleton } from "@/checkout-storefront/sections/DeliveryMethods/DeliveryMethodsSkeleton";
+import { useUser } from "@/checkout-storefront/hooks/useUser";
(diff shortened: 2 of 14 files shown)


packages/checkout-storefront/src/sections/Summary/useSummaryHeightCalc.ts
```diff
@@ -35,9 +39,7 @@ export const useSummaryHeightCalc = ({ linesCount, onBreakpointChange }: UseSumm
     };
 
     window.addEventListener("resize", handleWindowResize, { passive: true });
-    handleWindowResize();
-
-    return () => window.removeEventListener("resize", handleWindowResize);
```
mmiszy, line 40: This needs to stay – otherwise, we're adding a new event listener every time `[linesCount, onBreakpointChange]` changes but we never remove them.

mmiszy (changes requested):

packages/checkout-storefront/src/sections/DeliveryMethods/DeliveryMethods.tsx
```diff
@@ -38,26 +43,30 @@ export const DeliveryMethods: React.FC<CommonSectionProps> = ({ collapsed }) =>
...
-          {shippingMethods?.map(
-            ({ id, name, price, minimumDeliveryDays: min, maximumDeliveryDays: max }) => (
-              <SelectBox key={id} name="selectedMethodId" value={id}>
-                <div className="min-h-12 grow flex flex-col justify-center pointer-events-none">
-                  <div className="flex flex-row justify-between self-stretch items-center">
-                    <Text>{name}</Text>
-                    <Text>{getFormattedMoney(price)}</Text>
+        {authenticated && !shippingAddress && updateState.checkoutShippingUpdate ? (
+          <DeliveryMethodsSkeleton />
+        ) : (
```
witoszekdev, line 51: I'm confused, why are we showing skeleton when you're authenticated?

witoszekdev: if this part should be visible only when you're authenticated then maybe we should first wrap it in an if and then use the turnary operator, like this:

```ts
{authenticated && (
  updateState.checkoutShippingUpdate ? (
    // ...
  ) : (
    // ...
  )
)}
```

witoszekdev (approved):

packages/checkout-storefront/src/sections/DeliveryMethods/DeliveryMethods.tsx
bmigirl: Hm, the code you've suggested indicates that both the conditions should be displayed only when authenticated, notice that in the current code the last section will display no matter what so... I don't think so? Maybe naming that 3-part condition would make it clearer?

packages/checkout-storefront/src/sections/Summary/useSummaryHeightCalc.ts
bmigirl: Removed by mistake actually, changing

witoszekdev (approved):

packages/checkout-storefront/src/lib/auth/SaleorAuthClient.ts
```diff
@@ -81,15 +81,17 @@ export class SaleorAuthClient {
       const res: TokenRefreshResponse = await response.clone().json();
```
mmiszy, line 81: Let's make sure these updates are also applied to apps/storefront

mmiszy (changes requested):

witoszekdev (approved):

mmiszy (approved):

PR #771, opened 2023-03-09
Update Readme.md

mmiszy:


apps/storefront/lib/auth/Readme.md
```diff
@@ -68,7 +68,7 @@ const { isAuthenticating, signIn, signOut, checkoutSignOut } = useSaleorAuthCont
 
 #### **`useAuthChange({ onSignedIn, onSignedOut }: UseAuthChangeProps) => void`**
 
-Once the successful sign in or sign out happens, as well as failed authentication (for e.g when refresh token is expired and there's no way to obtain a new access token from the api) you'll probably want to trigger some actions. Hence we're provided a `useAuthChange` hook that'll listen to storage changes and trigger callbacks.
+Regardless of whether the sign-in or sign-out is successful, you probably want to react to it. Hence we provided a `useAuthChange` hook that listens to storage changes and triggers callbacks.
```
bmigirl, line 71: I don't think the `Regardless of whether sign in or sign out is successful` is accurate here considering that the hook won't run in case of an unsuccessful sign-out. It'll run if you:
- signed in successfully (onSignedIn)
- signed out successfully or your tokens expired and there is no way to refresh them (onSignedOut)

apps/storefront/lib/auth/Readme.md
```diff
@@ -194,16 +194,16 @@ export const App = () => {
...
 ```javascript
 const { apolloClient, resetClient } = useApolloClient(saleorAuthClient.fetchWithAuth);
 ```
 
-Becasue we're using the client to also retrieve unauthenticated data in SSR, we separated them into two - one for SSR, that can be used outside of React flow, and the other returned by our hook.
+Because we're using the client to also retrieve unauthenticated data in SSR, we separated them into two - one for SSR, which can be used outside of React flow, and the other returned by our hook.
 
 ```javascript
-export const serverApolloClient = new ApolloClient({
+export const staticApolloClient = new ApolloClient({
```
bmigirl, line 206: It's called `server` in the codebase, must have deviated while we're maintaining two codebases for this

packages/checkout-storefront/src/lib/auth/Readme.md
```diff
@@ -194,13 +194,13 @@ export const App = () => {
...
 
 ```javascript
 const { apolloClient, resetClient } = useApolloClient(saleorAuthClient.fetchWithAuth);
 ```
 
-Becasue we're using the client to also retrieve unauthenticated data in SSR, we separated them into two - one for SSR, that can be used outside of React flow, and the other returned by our hook.
+Because we're using the client to also retrieve unauthenticated data in SSR, we separated them into two - one for SSR, which can be used outside of React flow, and the other returned by our hook.
 
 ```javascript
 export const staticApolloClient = new ApolloClient({
```
bmigirl, line 206: Here also needs a change
```suggestion
export const serverApolloClient = new ApolloClient({
```

bmigirl (changes requested):

PR #833, opened 2023-05-09
Allow only certain api urls

mmiszy:


.env
```diff
@@ -1,4 +1,5 @@
 SALEOR_API_URL=https://master.staging.saleor.cloud/graphql/
+ALLOWED_SALEOR_API_REGEX="^https:\/\/[a-z0-9\.]+\.saleor\.cloud\/graphql\/\$"
```
witoszekdev, line 2: We should probably document this env variable in our docs - this only allows calls from `saleor.cloud` but if someone uses self-hosted version they should change this.

PR #867, opened 2023-07-06
Fix no cache user for user re-fetching

zaiste:


mmiszy (changes requested):
Could you also update `@saleor/auth-sdk` to the latest version?

apps/storefront/lib/useUser.ts
```diff
@@ -1,7 +1,7 @@
 import { useUserQuery } from "@/saleor/api";
 
 export const useUser = () => {
-  const { data, loading } = useUserQuery();
+  const { data, loading } = useUserQuery({ fetchPolicy: "no-cache" });
```
mmiszy, line 4: What about other queries that may depend on the user being logged in, for example `checkout`?

mmiszy (changes requested):

apps/storefront/lib/useUser.ts
zaiste: @mmiszy changed to `cache-and-network`

Dockerfile.base
```diff
@@ -2,7 +2,7 @@ FROM node:16-slim
 WORKDIR /app
 
 # Setup pnpm package manager
-RUN npm install -g pnpm@7.11.0
+RUN npm install -g pnpm
```
mmiszy, line 5: You should pin the version to avoid accidentally installing pnpm 9 and breaking the project

mmiszy (changes requested):

mmiszy (approved):

PR #883, opened 2023-09-14
Providing More Variable Configurations

saranshisatgit:
This allows 

1. Footer data 
2. Logo and SVG files and path that can be utilised


saranshisatgit:
I guess the PR was not full called for would there be a possible situation where those values could be dynamically set using ENV VAR?

PR #898, opened 2023-10-05
fix: fixes related to a11y

grzegorzpokorski:
- Improve accessibility of the VariantSelector (inc. replace buttons with radio inputs).
- Add missed headings and adjust existing ones.
- Disable link to checkout when there is no items in cart.

### DIFF
--- a/src/app/cart/CheckoutLink.tsx
+++ b/src/app/cart/CheckoutLink.tsx
@@ -1,6 +1,5 @@
 "use client";
 
-import clsx from "clsx";
 import Link from "next/link";
 
 type Props = {
@@ -14,10 +13,7 @@ export const CheckoutLink = ({ disabled, checkoutId }: Props) => {
 			aria-disabled={disabled}
 			onClick={(e) => disabled && e.preventDefault()}
 			href={`/checkout?checkout=${checkoutId}`}
-			className={clsx(
-				"aria-disabled:cursor-not-allowed",
-				"w-full rounded border border-transparent bg-pink-600 px-6 py-3 text-center font-medium text-gray-50 hover:bg-pink-500 aria-disabled:bg-pink-500 sm:col-start-2",
-			)}
+			className="w-full rounded border border-transparent bg-pink-600 px-6 py-3 text-center font-medium text-gray-50 hover:bg-pink-500 aria-disabled:cursor-not-allowed aria-disabled:bg-pink-500 sm:col-start-2"
 		>
 			Checkout
 		</Link>
--- a/src/app/cart/CheckoutLink.tsx
+++ b/src/app/cart/CheckoutLink.tsx
@@ -15,7 +15,7 @@ export const CheckoutLink = ({ disabled, checkoutId }: Props) => {
 			onClick={(e) => disabled && e.preventDefault()}
 			href={`/checkout?checkout=${checkoutId}`}
 			className={clsx(
-				disabled && "cursor-not-allowed",
+				"aria-disabled:cursor-not-allowed",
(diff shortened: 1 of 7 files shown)


src/app/cart/CheckoutLink.tsx
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export const CheckoutLink = ({ disabled, checkoutId }: Props) => {
+	return (
+		<Link
+			aria-disabled={disabled}
+			onClick={(e) => disabled && e.preventDefault()}
+			href={`/checkout?checkout=${checkoutId}`}
+			className={clsx(
+				disabled && "cursor-not-allowed",
```
mmiszy, line 18:
```suggestion
				"aria-disabled:cursor-not-allowed",
```
this should also work, right?

src/app/categories/[slug]/page.tsx
```diff
@@ -29,10 +29,11 @@ export default async function Page({ params }: { params: { slug: string } }) {
 		<div>
 			<div className="border-b bg-slate-100/50">
 				<div className="mx-auto max-w-7xl p-8">
-					<h2 className="text-xl font-semibold">{name}</h2>
+					<h1 className="text-xl font-semibold">{name}</h1>
 				</div>
 			</div>
 			<section className="sm:py-18 mx-auto max-w-2xl px-8 py-12 sm:px-6 lg:max-w-7xl">
+				<h2 className="sr-only">Product list</h2>
```
Comment (mmiszy) [inline src/app/categories/[slug]/page.tsx:36]: Maybe it should say something like "Products in category {name}"

src/ui/components/VariantSelector.tsx
```diff
@@ -39,28 +39,38 @@ export function VariantSelector(props: { variants: { id: string; name: string }[
 		}
 	}, [selectVariant, variants]);
 
+	const generatedId = useId();
+
 	return (
-		<div className="my-4">
-			<div className="sr-only">Variants</div>
+		<div className="my-4" role="radiogroup" aria-labelledby={`${generatedId}-title`}>
```
mmiszy, line 45: Maybe this should be a `fieldset` with `legend` instead? So we can avoid generating the id.

src/ui/components/VariantSelector.tsx
```diff
@@ -39,28 +39,38 @@ export function VariantSelector(props: { variants: { id: string; name: string }[
...
 					variants.map((variant) => {
 						return (
-							<button
+							<label
 								key={variant.id}
-								type="button"
-								onClick={() => {
-									selectVariant(variant.id);
-								}}
+								htmlFor={`${generatedId}-variant-${variant.id}`}
```
mmiszy, line 55: This is unnecessary if the input is inside the label element

mmiszy (changes requested):

src/ui/components/VariantSelector.tsx
grzegorzpokorski: Nice suggestion. I also considered to use that tag, but I wasn't sure if can I use it outside of the form.

grzegorzpokorski (commented):
After careful consideration and testing of keyboard navigation between variants, I realized that when an option is selected in a group, the focus also shifts along with the selection. This causes the page to rerender every time the user navigates through options in the group, due to a change in the variant in the search parameters. Because of this, I reverted to using buttons with appropriate ARIA roles. While this ensures that buttons in the group are correctly announced by screen readers, moving focus between options no longer triggers a page rerender - the selection doesn't follow the focus. I found this solution implemented in some online stores, such as the Logitech store ([example](https://www.logitech.com/pl-pl/products/mice/mx-master-3s.910-006560.html)).

zaiste (approved):

zaiste:
@grzegorzpokorski Fantastic changes overall! Thanks for your contributions

PR #912, opened 2023-10-09
fix: fix tile size

### LINKED ISSUE #911: Tiles vary in size
### Summary

Consider using `object-fit` to place images within constant ratio tiles, which would also have positive effect on layout shift
image

### Steps to reproduce

Use different image proportions

_No response_

### Output from envinfo

_No response_

grzegorzpokorski:
FIx #911 issue.

Added additional props to be set in the `/ui/atoms/Image` component:
- `fit` to determine how the image should be fitted inside the parent container,
- `square` to specify whether we want to display the image in a square aspect ratio.

Below are examples of how it looks in some configurations:
- on product list:
object-fit-variants
- on product page:
Zrzut ekranu 2023-10-09 190912

### DIFF
--- a/src/app/products/[slug]/page.tsx
+++ b/src/app/products/[slug]/page.tsx
@@ -110,7 +110,17 @@ export default async function Page(props: { params: { slug: string }; searchPara
 	return (
 		<section className="mx-auto grid max-w-7xl p-8">
 			<form className="grid gap-2 sm:grid-cols-2" action={addItem}>
-				{firstImage && <Image alt={firstImage.alt ?? ""} width={1024} height={1024} src={firstImage.url} />}
+				{firstImage && (
+					<div className="">
+						<Image
+							alt={firstImage.alt ?? ""}
+							width={1024}
+							height={1024}
+							src={firstImage.url}
+							contain="object-contain"
+						/>
+					</div>
+				)}
 				<div className="flex flex-col justify-between pt-6 sm:px-6 sm:pt-0">
 					<div>
 						<h1 className="flex-auto text-3xl font-bold tracking-tight text-slate-900">{product?.name}</h1>
--- a/src/app/products/[slug]/page.tsx
+++ b/src/app/products/[slug]/page.tsx
@@ -117,7 +117,7 @@ export default async function Page(props: { params: { slug: string }; searchPara
 							width={1024}
 							height={1024}
 							src={firstImage.url}
-							contain="object-contain"
+							fit="object-contain"
 						/>
 					</div>
 				)}
--- a/src/ui/atoms/Image.tsx
+++ b/src/ui/atoms/Image.tsx
@@ -1,3 +1,4 @@
+import clsx from "clsx";
 import NextImage from "next/image";
 
 type Props = {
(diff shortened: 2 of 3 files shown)


src/ui/atoms/Image.tsx
```diff
@@ -1,21 +1,28 @@
+import clsx from "clsx";
 import NextImage from "next/image";
 
 type Props = {
 	src: string;
 	alt: string;
 	width: number;
 	height: number;
+	square?: boolean;
```
mmiszy, line 9: Is this used anywhere?

mmiszy, line 9: Is this used anywhere?

src/ui/atoms/Image.tsx
```diff
@@ -1,21 +1,28 @@
...
+	fit?: "object-contain" | "object-cover";
```
mmiszy, line 10: Is this used anywhere with `object-cover`?

src/app/products/[slug]/page.tsx
```diff
@@ -110,7 +110,17 @@ export default async function Page(props: { params: { slug: string }; searchPara
 	return (
 		<section className="mx-auto grid max-w-7xl p-8">
 			<form className="grid gap-2 sm:grid-cols-2" action={addItem}>
-				{firstImage && <Image alt={firstImage.alt ?? ""} width={1024} height={1024} src={firstImage.url} />}
+				{firstImage && (
+					<div className="">
```
Comment (mmiszy) [inline src/app/products/[slug]/page.tsx:114]: What does this div do?

mmiszy (changes requested):

src/app/products/[slug]/page.tsx
Comment (grzegorzpokorski) [reply src/app/products/[slug]/page.tsx:114]: You are right, it is redundant 👍

src/ui/atoms/Image.tsx
grzegorzpokorski: Right now nowhere. I just wanted to give a posibility to choose between this two options. Sometimes, it looks good when images are displayed in a container with `object-fit: cover;` but on the product page, we want to see the entire product image, so the ideal choice is to display the image with `object-fit: contain;` - you can look at the examples in PR description. Maybe I add this props exaggeratedly and I should leave hardcoded only one option, I think `object-fit: contain;` should be the best. What do you think?

src/ui/atoms/Image.tsx
grzegorzpokorski: Maybe I should hardcode aspect ratio.

src/ui/atoms/Image.tsx
mmiszy: This component has been now modified and accepts any props

grzegorzpokorski: Ok, after consideration, I've opted to stick with `object-fit: contain;` because, from the user's perspective, it's crucial to see the entire product, which may not always be possible when the image has the `object-fit: cover;` property. 
If there is a need to present the image in a different manner, then it will be an appropriate moment to introduce additional options to the component responsible for displaying the image.

grzegorzpokorski: @mmiszy sorry, I didn't see your comment because I had the stale page open.
Newly created component that receive any props have a default set of classes. So, should I extend them or override these classes in the places where we need to resolve this issue? In your opinion, which approach will be better? Now I simply modified default set of classes. Let me know if shoul I change it.

mmiszy (approved):

PR #928, opened 2023-10-12
fix: improve performance and styles

grzegorzpokorski:
- Unify style of the product lists in app - grid breakpoints, spacings and widths.
- Add the `sizes` attribute to the images in the product list to provide the browser with more information about how much space they occupy on the screen - it should improve overall lighthouse performance score.
- Add the `priority` property to the thumbnail on product page because it is LCP element on that page.
- Add missed margin in cart.

### DIFF
--- a/src/app/(main)/page.tsx
+++ b/src/app/(main)/page.tsx
@@ -1,6 +1,6 @@
 import { ProductListDocument } from "@/gql/graphql";
 import { execute } from "@/lib/graphql";
-import { ProductElement } from "@/ui/components/ProductElement";
+import { ProductList } from "@/ui/components/ProductList";
 
 export const metadata = {
 	title: "Saleor Storefront example",
@@ -20,10 +20,8 @@ export default async function Page() {
 			{/* <CollectionList /> */}
 			<section className="mx-auto max-w-7xl p-8">
 				<h2 className="sr-only">Product list</h2>
-				<div className="mt-4 grid gap-8 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4">
-					{products.map((product, index) => (
-						<ProductElement key={product.id} product={product} loading={index < 4 ? "eager" : "lazy"} />
-					))}
+				<div className="mt-4">
+					<ProductList products={products} />
 				</div>
 			</section>
 		</div>
--- a/src/ui/components/ProductList.tsx
+++ b/src/ui/components/ProductList.tsx
@@ -1,0 +1,16 @@
+import { ProductElement } from "./ProductElement";
+import { type ProductFragment } from "@/gql/graphql";
+
+type Props = {
+	products: ProductFragment[];
+};
+
+export const ProductList = ({ products }: Props) => {
+	return (
+		<ul className="grid gap-8 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4" role="list">
+			{products.map((product, index) => (
+				<ProductElement key={product.id} product={product} loading={index < 4 ? "eager" : "lazy"} />
+			))}
+		</ul>
+	);
+};
--- a/src/ui/components/ProductList.tsx
+++ b/src/ui/components/ProductList.tsx
@@ -9,7 +9,12 @@ export const ProductList = ({ products }: Props) => {
 	return (
 		<ul className="grid gap-8 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4" role="list">
 			{products.map((product, index) => (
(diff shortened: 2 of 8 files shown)


zaiste (approved):

src/app/(main)/page.tsx
```diff
@@ -20,10 +20,8 @@ export default async function Page() {
 			{/* <CollectionList /> */}
 			<section className="mx-auto max-w-7xl p-8">
 				<h2 className="sr-only">Product list</h2>
-				<div className="mt-4 grid gap-8 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4">
-					{products.map((product, index) => (
-						<ProductElement key={product.id} product={product} loading={index < 4 ? "eager" : "lazy"} />
-					))}
+				<div className="mt-4">
```
mmiszy, line 23: Why is it `mt-4` here but `my-8` in other places?

mmiszy, line 23: Why is it `mt-4` here but `my-8` in other places?

src/app/(main)/page.tsx
```diff
@@ -20,10 +20,8 @@ export default async function Page() {
 			{/* <CollectionList /> */}
 			<section className="mx-auto max-w-7xl p-8">
 				<h2 className="sr-only">Product list</h2>
-				<div className="mt-4 grid gap-8 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4">
-					{products.map((product, index) => (
-						<ProductElement key={product.id} product={product} loading={index < 4 ? "eager" : "lazy"} />
```
mmiszy, line 25: Looks like most of the code in this PR is a result of a bad rebase?

mmiszy, line 25: Looks like most of the code in this PR is a result of a bad rebase?

src/ui/components/ProductList.tsx
```diff
@@ -0,0 +1,21 @@
+import { ProductElement } from "./ProductElement";
```
mmiszy, line 1: This file is already on `canary` so how come it shows as a completely new here?
Bad rebase?

src/ui/components/ProductList.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const ProductList = ({ products }: Props) => {
+	return (
+		<ul className="grid gap-8 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4" role="list">
+			{products.map((product, index) => (
+				<ProductElement
+					key={product.id}
+					product={product}
+					loading={index < 4 ? "eager" : "lazy"}
+					sizes="(max-width: 640px) 100vw, (max-width: 768px) 50vw, (max-width: 1024px) 33vw, 25vw"
```
mmiszy, line 16: Where do these values come from?
What are we trying to achieve here, given that the performance score in Lighthouse is already 100/100?

grzegorzpokorski: @mmiszy Great questions!

The values referenced in the media queries in sizes attribute are derived from Tailwind CSS breakpoints, serving as the breakpoints used by it - and I know It looks little messy, but I couldn't found better way to provide them here in this case. The second argument in these media queries depends on the grid template applied to the parent ul element - for instance, `grid-cols-3` => `33vw`.

But I why even add this it here? What I want achive? As you mentioned, the application achieves a high Lighthouse score, often reaching 100/100. However, there is recurring feedback indicating that image sizes could be better adjusted, particularly during mobile device testing.

pagespeed web dev_analysis_https-storefront-saleor-example-vercel-app_vi1ifnwc8x_form_factor=mobile

I believe that providing the sizes attribute will contribute positively to this metric. While the `srcset` attribute, automatically provided by next/image, allows the browser to decide which image to load based on pixel density, the `sizes` attribute offers additional information about the real space an image will occupy.

I might be wrong. Maybe it is unnecessary and present results are good enough.

mmiszy: Okay, but grid-cols-3 is not exactly 33vw, right?

grzegorzpokorski: > Okay, but grid-cols-3 is not exactly 33vw, right?

Yeah, that's true. I think it isn't possible to provide a exact value of space which image will occupy since we can only pass a value in `vw` units. I tested results of this PR in Lighthouse and it still notify about possibility to shrink images. despite this in the end it reduced amount of data required to load page. Maybe it's not worth sacrificing code readability 🤔

PR #935, opened 2023-10-12
Adjust colors

timuric:


mmiszy:
How is it different from https://github.com/saleor/storefront/pull/934 ?

timuric:
> How is it different from #934 ?

This includes changes to navigation header, but you don't have to merge it, just cherry pick them
image

PR #951, opened 2023-10-13
Remove redundant column in footer

timuric:


zaiste:
@timuric we can remove the `footer` variable as well

zaiste (approved):

PR #978, opened 2023-10-19
fix: improve a11y

grzegorzpokorski:


src/app/(main)/products/page.tsx
```diff
@@ -32,6 +32,7 @@ export default async function Page({ searchParams }: Props) {
 
 	return (
 		<div>
+			<h1 className="sr-only">All products</h1>
 			<section className="mx-auto max-w-7xl p-8 pb-16">
 				<h2 className="sr-only">Product list</h2>
```
mmiszy, line 37: Is this useful is `All products` is already in h1?
So the page reads `All products Product list`.

mmiszy (changes requested):

src/app/(main)/products/page.tsx
grzegorzpokorski: @mmiszy Good point. I missed to test it. And I forget a real purpose of this headings. The change has been reversed.

mmiszy (approved):

PR #979, opened 2023-10-20
Go back to store button is not working

### LINKED ISSUE #903: Go back to store button is not working
### Summary

There is no action with "Go back to store " button

### Steps to reproduce

1. Add 3 items to the basket
2. Go to checkout
3. In checkout set quantity to 0 for all items
Current result: You will see a popup with button 'go back to store'
4. Click on 'go back to store'
Result: Nothing happen, there is no action with this button. What is more, the basket icon shows that there are still 2 items in the basket.
image

_No response_

### Output from envinfo

_No response_

KriteshTimsina:
fixed #903


mmiszy:
Thanks for this PR! Per our design principles, we don't use next.js-specific components in `src/checkout`.

https://github.com/saleor/storefront/blob/e29ebbf7b20d9c3dcf710dd715880fd414a3ad2f/README.md?plain=1#L53

mmiszy (changes requested):
We can replace this with a simple `<a href={process.env.NEXT_PUBLIC_STOREFRONT_URL || "/"}>`

zaiste:
Fixed in #993

PR #1010, opened 2023-11-09
Add update types GA workflow

2can:
This workflow checks if there is a new release of Saleor with the attached `schema.graphql` file. If the file is found it runs  `pnpm generate` to update types and issue a PR.


.github/workflows/update_types.yml
```diff
@@ -0,0 +1,75 @@
...
+  schedule:
+    - cron: "0 18 * * 1-5" # workdays at 18:00
+
+jobs:
+  check_saleor_release:
+    name: Check Saleor release and update types
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
mmiszy, line 14: Can we move this step after checking if there's a new release? Should be faster

.github/workflows/update_types.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Download schema.graphql
+        run: |
+          if [[ "${{ steps.release.outputs.has_schema }}" == "true" ]]; then
+            echo "schema.graphql is attached to the ${{ steps.release.outputs.tag }} release, updating..."
+          else
+            echo "schema.graphql is not attached to the ${{ steps.release.outputs.tag }} release, exiting..."
+            exit 1
+          fi
+
+      - uses: robinraju/release-downloader@v1.8
```
mmiszy, line 33: We could probably not use this action and just use CURL do download it? It's a single file and we already have the path from the previous steps.

mmiszy (changes requested):

.github/workflows/update_types.yml
2can: Done

.github/workflows/update_types.yml
2can: Done

mmiszy (approved):

mmiszy (approved):

PR #1011, opened 2023-11-10
Add order list for logged user

zaiste:


mmiszy:
Could you rebase @zaiste ?

mmiszy (changes requested):

src/app/(main)/orders/page.tsx
```diff
@@ -60,7 +56,7 @@ export default function OrderPage() {
 											</div>
 											<div>
 												<dt className="font-medium text-slate-900">Total amount</dt>
-												<dd className="mt-1 font-medium text-slate-900">{0}</dd>
+												<dd className="mt-1 font-medium text-slate-900">{order.total?.gross.amount}</dd>
```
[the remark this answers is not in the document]
mmiszy: Why `text-slate-900` instead of `text-neutral-900` ?

mmiszy (changes requested):
Replace `slate` with `neutral`

mmiszy:
This doesn't look too good:

Screenshot 2023-11-10 at 17 09 26

package.json
```diff
@@ -16,18 +16,19 @@
...
 	"dependencies": {
 		"@adyen/adyen-web": "5.53.2",
-		"@adyen/api-library": "14.3.0",
+		"@adyen/api-library": "14.4.0",
 		"@headlessui/react": "1.7.17",
 		"@saleor/auth-sdk": "0.14.0",
 		"@stripe/react-stripe-js": "2.3.1",
 		"@stripe/stripe-js": "2.1.11",
 		"clsx": "2.0.0",
+		"date-fns": "2.30.0",
```
mmiszy, line 25: We don't need a library to format a date. We can just use `Intl`:
```
new Intl.DateTimeFormat('en-US').format(date)
```

src/app/(main)/orders/page.tsx
```diff
@@ -0,0 +1,7 @@
+import dynamic from "next/dynamic";
+
+export default function OrderPage() {
+	const OrderList = dynamic(() => import("../../../ui/components/OrderList"), { ssr: false });
```
mmiszy, line 4: I think this should be outside the component body?
```suggestion
const OrderList = dynamic(() => import("../../../ui/components/OrderList").then((m) => m.OrderList), { ssr: false });

export default function OrderPage() {
```

src/ui/components/OrderList.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { CurrentUserOrderListDocument, type CurrentUserOrderListQuery } from "@/gql/graphql";
+import { formatMoney } from "@/lib/graphql";
+
+// eslint-disable-next-line import/no-default-export
+export default function OrderList() {
+	const [{ data, fetching, error }] = useQuery<CurrentUserOrderListQuery>({
+		query: CurrentUserOrderListDocument.toString(),
+	});
+
+	console.log("here we go", fetching, data, error);
```
mmiszy, line 14:
```suggestion
```

src/ui/components/OrderList.tsx
```diff
@@ -0,0 +1,117 @@
+"use client";
+import { useQuery } from "urql";
+import Image from "next/image";
+import { format } from "date-fns";
+import { CurrentUserOrderListDocument, type CurrentUserOrderListQuery } from "@/gql/graphql";
+import { formatMoney } from "@/lib/graphql";
+
+// eslint-disable-next-line import/no-default-export
+export default function OrderList() {
```
mmiszy, line 9:
```suggestion
export function OrderList() {
```

src/ui/components/OrderList.tsx
```diff
@@ -0,0 +1,117 @@
...
+		query: CurrentUserOrderListDocument.toString(),
+	});
+
+	console.log("here we go", fetching, data, error);
+
+	if (fetching) {
+		return <div>Loading...</div>;
+	}
+
+	if (!fetching) {
```
mmiszy, line 20: This condition is redundant because line 16 makes it impossible not to be true

mmiszy (changes requested):

src/app/(main)/orders/page.tsx
zaiste: @mmiszy named export doesn't work unfortunately

mmiszy: It does work, we're already using in other places.

zaiste: @mmiszy can you check on your side?

CleanShot 2023-11-11 at 14 00 10@2x

(with re-created `.next/`)

mmiszy: Indeed there's a bug in Next.js :D Adding "use client" to the `page.tsx` fixes it.

zaiste: :D 

Yeah, but I'd prefer not to set the entire page as `client`. Let's keep it as `default` for now with an eslint exception, there's an open bug report in the Next.js repo, let's track it and fix it once it's done upstream. WDYT?

mmiszy (approved):

src/app/(main)/orders/page.tsx
mmiszy: This whole page is client anyway because you're only rendering a single dynamic component.

PR #1014, opened 2023-11-11
feat: add user menu to the navbar

grzegorzpokorski:
Changes:
1. Fixed user avatar by providing a placeholder built from user's first & last name or email address.
2. Added a user menu to the navbar, which contains links related to the user profile and a button to log out. This is a more opinionated approach to presenting user-related links.
3. Improved the title of the orders page by providing an user's email address as a fallback value for the user's first name.
4. Redirect user after logging to the home page (`/login` page are reachable only for non logged users).
5. Remove UserCard component - it is no longer needed.

What issues resolve?
1. The user can only log out by clicking a button on the /login path, but there is no link to that page when the user is logged in.
2. The user may not have an avatar.
3. The user may not have a first and/or last name.

Before:

przed

After:
1. When the user doesn't have an avatar and doesn't have defined first name (email as a fallback):

po2

6. When the user has an avatar and defined first & last name :

localhost_3000_orders


zaiste (dismissed):

src/ui/components/nav/components/UserMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+							width={24}
+							height={24}
+							alt=""
+						/>
+					) : (
+						<span
+							className="flex h-8 w-8 items-center justify-center rounded-full border bg-white text-center text-xs font-bold uppercase"
+							aria-hidden="true"
+						>
+							{data.me.email.slice(0, 2)}
```
mmiszy, line 38: We typically use initials for this. This could be a separate component because the logic quickly becomes complicated

src/ui/components/nav/components/UserMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+										My orders
+									</Link>
+								)}
+							</Menu.Item>
+						</div>
+						<div className="flex flex-col px-1 py-1">
+							<Menu.Item>
+								{({ active }) => (
+									<button
+										onClick={() => signOut()}
```
mmiszy, line 72:
```suggestion
										type="button"
										onClick={() => signOut()}
```

mmiszy (changes requested):

zaiste (approved):

PR #1019, opened 2023-11-13
feat: add json-ld and individual og image for single product page

grzegorzpokorski:
- Change the OG image for the product page to the product image instead of the default OG image.
- Add JSON-LD for product pages.

### DIFF
--- a/src/app/(main)/products/[slug]/page.tsx
+++ b/src/app/(main)/products/[slug]/page.tsx
@@ -5,6 +5,7 @@ import { notFound } from "next/navigation";
 import { type ResolvingMetadata, type Metadata } from "next";
 import xss from "xss";
 import invariant from "ts-invariant";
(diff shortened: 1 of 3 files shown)


src/app/(main)/products/[slug]/page.tsx
```diff
@@ -125,8 +134,46 @@ export default async function Page(props: { params: { slug: string }; searchPara
 		  })
 		: "";
 
+	const productJsonLd = {
```
Comment (mmiszy) [inline src/app/(main)/products/[slug]/page.tsx:137]: You can install a package `schema-dts` and then add types to this object:
```ts
import { type Product } from "schema-dts";
```

mmiszy (changes requested):

mmiszy (approved):

mmiszy (approved):

PR #1032, opened 2023-11-16
Feature: Added searchBar

aidensgithub:
## Added a simple SearchBar
Styled in Tailwind and the Search icon was taken from [Lucide](https://lucide.dev/icons/search)

**View on home screen**

image
---
**View on search results**

image
---
**View on no results**

image
---
**View on mobile**

image

### DIFF
--- a/src/ui/components/nav/components/SearchBar.tsx
+++ b/src/ui/components/nav/components/SearchBar.tsx
@@ -1,0 +1,56 @@
+"use client";
+
+import { useRouter, useSearchParams, type ReadonlyURLSearchParams } from "next/navigation";
+import { useEffect, useState } from "react";
+import { Search } from "lucide-react";
+
+export const createUrl = (pathname: string, params: URLSearchParams | ReadonlyURLSearchParams) => {
+	const paramsString = params.toString();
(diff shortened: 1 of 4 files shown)


src/ui/components/nav/components/SearchBar.tsx
```diff
@@ -0,0 +1,56 @@
+"use client";
```
mmiszy, line 1: I think this doesn't need to be a client component. You can use a server component with [server actions](https://nextjs.org/docs/app/api-reference/functions/server-actions) and call [`redirect`](https://nextjs.org/docs/app/api-reference/functions/redirect)

mmiszy (changes requested):
Thank you so much for this contribution! I left a suggestion.

zaiste:
@aidensgithub pretty cool addition!

src/ui/components/nav/components/SearchBar.tsx
zaiste: @aidensgithub yes, it's more idiomatic in this context (App Router)

zaiste (dismissed):

zaiste (approved):

PR #1035, opened 2023-11-18
fix/refactor: refactor search form and pagination

grzegorzpokorski:
Changes / fixes:
1. Prevent submission of an empty search form.
2. Enhance the visibility and accessibility of the search form (add a background to the text input, include labels, implement built-in form validation).
3. Improve the types of `searchParams`.
4. Close the mobile menu when the search parameters in the URL change (e.g. "query" string).
5. Handle the case when there is no `query` parameter in the URL on the `/search` page or there is more than one search param named `query`.
6. Make the `Pagination` component more reusable:
6.1. introduce `basePathname` prop (instead of hardcoding it in the component),
6.2. introduce `urlSearchParams` prop.
7. Reduced layout shift in header after login/out.

### DIFF
--- a/src/app/(main)/products/page.tsx
+++ b/src/app/(main)/products/page.tsx
@@ -3,6 +3,7 @@ import { ProductListPaginatedDocument } from "@/gql/graphql";
 import { ProductsPerPage, executeGraphQL } from "@/lib/graphql";
 import { Pagination } from "@/ui/components/Pagination";
 import { ProductList } from "@/ui/components/ProductList";
+import { parseCursor } from "@/lib/utils";
 
 export const metadata = {
 	title: "Products · Saleor Storefront example",
@@ -11,12 +12,12 @@ export const metadata = {
 
 type Props = {
 	searchParams: {
-		cursor: string;
+		cursor: string | string[] | undefined;
 	};
 };
 
 export default async function Page({ searchParams }: Props) {
-	const { cursor } = searchParams;
+	const cursor = parseCursor(searchParams.cursor);
 
 	const { products } = await executeGraphQL(ProductListPaginatedDocument, {
 		variables: {
@@ -30,6 +31,10 @@ export default async function Page({ searchParams }: Props) {
 		notFound();
 	}
 
+	const newSearchParams = new URLSearchParams({
+		...(products.pageInfo.endCursor && { cursor: products.pageInfo.endCursor }),
+	});
+
 	return (
 		<div>
 			<section className="mx-auto max-w-7xl p-8 pb-16">
@@ -39,7 +44,7 @@ export default async function Page({ searchParams }: Props) {
 					pageInfo={{
 						...products.pageInfo,
 						basePathname: "/products",
-						urlSearchParams: new URLSearchParams(searchParams),
+						urlSearchParams: newSearchParams,
 					}}
 				/>
 			</section>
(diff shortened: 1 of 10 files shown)


src/ui/components/Pagination.tsx
```diff
@@ -5,14 +5,20 @@ export async function Pagination({
 	pageInfo,
 }: {
 	pageInfo: {
-		endCursor?: string | null;
+		baseUrl: string;
 		hasNextPage: boolean;
+		endCursor?: string | null;
+		urlSearchParams?: URLSearchParams;
```
mmiszy, line 11:
```suggestion
		urlSearchParams?: ReadonlyURLSearchParams;
```

src/app/(main)/search/page.tsx
```diff
@@ -7,11 +8,19 @@ export const metadata = {
...
-	searchParams?: { [key: string]: string | string[] | undefined };
+	searchParams?: Record<string, string>;
 };
-export default async function Page({ searchParams }: { searchParams?: Props }) {
-	const { cursor, query: searchValue } = searchParams as { [key: string]: string };
+
+export default async function Page({ searchParams }: Props) {
+	const urlSearchParams = new URLSearchParams(searchParams);
+
+	if (!urlSearchParams.has("query")) notFound();
```
mmiszy, line 19:
```suggestion
	if (!urlSearchParams.has("query")) {
		notFound();
	}
```

src/ui/components/Pagination.tsx
```diff
@@ -5,14 +5,20 @@ export async function Pagination({
...
 	};
 }) {
+	if (pageInfo.endCursor) {
+		pageInfo.urlSearchParams?.set("cursor", pageInfo.endCursor);
```
mmiszy, line 15: This line mutates `pageInfo.urlSearchParams` – a parameter coming from the props. That can lead to unexpected results.

src/app/(main)/products/page.tsx
```diff
@@ -35,7 +35,13 @@ export default async function Page({ searchParams }: Props) {
 			<section className="mx-auto max-w-7xl p-8 pb-16">
 				<h2 className="sr-only">Product list</h2>
 				<ProductList products={products.edges.map((e) => e.node)} />
-				<Pagination pageInfo={products.pageInfo} />
+				<Pagination
+					pageInfo={{
+						...products.pageInfo,
+						baseUrl: "/products",
```
mmiszy, line 41: It's not a full URL, it's just a pathname. Let's rename it to `basePathname`

mmiszy (changes requested):

src/ui/components/Pagination.tsx
grzegorzpokorski: I am not sure but `ReadonlyURLSearchParams` is a internal class used by next.js e.g. in `useSearchParams` hook and it shouldn't be used here. Maybe insted I change it to `readonly urlSearchParams?: URLSearchParams;`?

mmiszy: `import { ReadonlyURLSearchParams } from "next/navigation";`

grzegorzpokorski: For me it's look that result type of `new URLSearchParams()` is not compatible with type `ReadonlyURLSearchParams` because I receive an type error: `Property '[INTERNAL_URLSEARCHPARAMS_INSTANCE]' is missing in type 'URLSearchParams' but required in type 'ReadonlyURLSearchParams'.`

mmiszy: Hmmm yeah you're right, I didn't notice that. We should have our own Readonly… type for that. Let's ignore it for now.

zaiste (approved):

PR #1040, opened 2023-11-23
Migrate Auth to App Router

zaiste:


src/lib/auth.ts
```diff
@@ -0,0 +1,13 @@
+import { createSaleorAuthClient } from "@saleor/auth-sdk";
+import { getNextServerCookiesStorage } from "@saleor/auth-sdk/next/server";
+import { invariant } from "ts-invariant";
+
+const nextServerCookiesStorage = getNextServerCookiesStorage();
+
+invariant(process.env.NEXT_PUBLIC_SALEOR_API_URL, "Missing NEXT_PUBLIC_SALEOR_API_URL env variable");
+
+export const authClient = createSaleorAuthClient({
```
mmiszy, line 9: We could reuse the instance created in graphql.ts and then delete thus file

zaiste: @mmiszy yeah, I wanted to move `<Provider>` into checkout's `layout.tsx` and clearly separate things

mmiszy (approved):

src/app/config.ts
```diff
@@ -0,0 +1,15 @@
+import { createSaleorAuthClient } from "@saleor/auth-sdk";
+import { getNextServerCookiesStorage } from "@saleor/auth-sdk/next/server";
+import { invariant } from "ts-invariant";
+
+export const ProductsPerPage = 12;
+
+const saleorApiUrl = process.env.SALEOR_API_URL;
+invariant(saleorApiUrl, "Missing SALEOR_API_URL env variable");
```
mmiszy, line 8:
```suggestion
const saleorApiUrl = process.env.NEXT_PUBLIC_SALEOR_API_URL;
invariant(saleorApiUrl, "Missing NEXT_PUBLIC_SALEOR_API_URL env variable");
```

mmiszy (changes requested):

src/lib/auth.ts
mmiszy: Is this done?

src/lib/auth.ts
mmiszy, line 1: This file and `src/app/config.ts` are very similar, so we need both?

mmiszy (changes requested):

mmiszy (approved):

PR #1046, opened 2023-11-27
Add channel slug to URLs

### LINKED ISSUE #1029: Pass current channel in URL
Channel should prefix all other product-related routes, i.e. `/default-channel/products` instead of just `/products`.

There are multiple channels to be fetched from the API, however, first step of this issue might be to hardcode the `default-channel`.

mmiszy:
Fixes #1029


zaiste:
@mmiszy `default-channel` doesn't have to be visible in the pathname.

grzegorzpokorski (commented):
I'm not sure, but we should also add a channel in the redirectins on the search page and `SearchBar` component.

grzegorzpokorski:
@mmiszy CheckoutLink on the /cart page also lacks a channel in the href. Perhaps a good solution would be a middleware that adds a default channel to the links in situations where it doesn't have a locale in the pathname?

mmiszy:
@grzegorzpokorski `CheckoutLink` doesn't need a channel.

Issue #1024, opened 2023-11-15
Auto capitalize postal code


timuric:
Backend API is case sensitive, which creates unnecessary friction for the user, and sometimes it can be confusing why lower case postal code is invalid

kisharnath:
Hey I want to work on this issue

theRaven-code:
Hey I want to work on this issue, if it is not resolved yet

PR #1071, opened 2024-01-31
Auto capitalize postal code #1024

theRaven-code:
I've enforced the type to secure the postalCode input as Uppercase by…


peelar:
Hi @theRaven-code, sorry for long time with no answer. Thanks for your contribution. Unfortunately, this won't work because the file you edited is generated automatically. With a change in schema, it will regenerate and your changes won't be applied. This has to be enforced in runtime, but through types. I am closing this one.

PR #1104, opened 2024-09-10
Enable payment with dummy payment app

delemeator:
This is a quick and dirty cherry pick from my playground storefront to use payments from dummy payment app

Not tested - so use with caution


peelar:
https://github.com/saleor/storefront/pull/1115

PR #1114, opened 2025-02-05
Bug: Graphql codegen issue in checkout folder

krunal-ss:
### Description

- In the **storefront**, the GraphQL code generation (`codegen`) configuration was previously set to look for files only in the `src/graphql/` folder.  
- As a result, it excluded files from `src/checkout/graphql/*.graphql` and worked only for the `src/app` folder.  

### Solution  
- The `codegen` configuration will now also look into the `src/checkout/graphql/` folder.  
- It will execute the `codegen.yml` file and generate/modify the `index.ts` file in `src/checkout/graphql/`.  
- Additionally, it will create a `schema.graphql.json` file.


package.json
```diff
@@ -3,28 +3,31 @@
 	"version": "0.1.0",
 	"private": true,
 	"type": "module",
-	"license": "BSD-3-Clause",
```
krzysztofzuraw, line 6: please don't remove license from package.json 😄

package.json
```diff
@@ -3,28 +3,31 @@
...
 	"scripts": {
 		"dev": "next dev",
 		"build": "next build",
 		"start": "next start",
 		"predev": "pnpm run generate",
 		"prebuild": "pnpm run generate",
 		"lint": "next lint --dir src --fix",
-		"generate": "graphql-codegen --config .graphqlrc.ts",
+		"generate": "DOTENV_CONFIG_PATH=.env graphql-codegen -r dotenv/config --config ./src/checkout/codegen.yml && graphql-codegen --config .graphqlrc.ts",
```
krzysztofzuraw, line 13: **Suggestion:** I would rather see you add additional config entry to existing `.graphqlrc.ts` instead of creating new config file.

package.json
```diff
@@ -3,28 +3,31 @@
...
 		"prepare": "husky install",
 		"test": "playwright test"
 	},
 	"dependencies": {
 		"@adyen/adyen-web": "5.53.3",
 		"@adyen/api-library": "15.0.0-beta",
 		"@headlessui/react": "1.7.18",
+		"@hookform/resolvers": "3.10.0",
```
krzysztofzuraw, line 21: **Question:** why those new dependencies are needed?

kzuraw (changes requested):

package.json
krunal-ss: Fixed  😅

package.json
krunal-ss: Added a new config. in `.graphqlrc.ts` and remove `codegen.yml` file

package.json
krunal-ss: Remove unwanted dependency as I was working with some new functionality.

.graphqlrc.ts
```diff
@@ -20,8 +20,46 @@ if (!schemaUrl) {
 const config: CodegenConfig = {
 	overwrite: true,
 	schema: schemaUrl,
-	documents: "src/graphql/**/*.graphql",
+	documents: "src/**/*.graphql",
 	generates: {
+		"src/checkout/graphql/index.ts": {
+			plugins: [
+				{
+					add: {
```
krzysztofzuraw, line 28: **Question:** why do you need to add this?

```ts
// THIS FILE IS GENERATED WITH `pnpm generate`
import "graphql/language/ast"
```

PR #1115, opened 2025-02-24
Add support for dummy payment gateway

krzysztofwolski:
Use https://github.com/saleor/dummy-payment-app/tree/main  for testing and demo purposes

### DIFF
--- a/src/checkout/sections/PaymentSection/utils.ts
+++ b/src/checkout/sections/PaymentSection/utils.ts
@@ -24,14 +24,6 @@ export const getFilteredPaymentGateways = (
 	// we want to use only payment apps, not plugins
 	return compact(paymentGateways).filter(({ id, name }) => {
 		const shouldBeIncluded = supportedPaymentGateways.includes(id);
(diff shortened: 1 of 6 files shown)


src/checkout/sections/PaymentSection/utils.ts
```diff
@@ -24,14 +24,6 @@ export const getFilteredPaymentGateways = (
 	// we want to use only payment apps, not plugins
 	return compact(paymentGateways).filter(({ id, name }) => {
 		const shouldBeIncluded = supportedPaymentGateways.includes(id);
```
krzysztofwolski, line 26: Removed this if statement because it was redundant - only the allow-listed gateways will be returned. Also it caused issue with our dummy gateway - its id starts with saleor.io 🙃

witoszekdev (approved):
Very nice 🫶

krzysztofzuraw (approved):

korycins:
Shouldn't you also mention the dummy-payment gateway in the Readme? Currently it could be misleading, as readme says that only adyenapp is supported:
image

PR #1210, opened 2026-06-08
Paper Checkout v2

mirekm:
Paper Checkout v2 is now aligned with Next.js 16 App Router: server-owned cart state, server actions for Saleor mutations, BFF session shared with storefront, URL-driven steps without full page reloads, and a dedicated confirmation route - with an extensible payment gateway registry and first-class Stripe + Dummy integrations.


src/checkout/lib/payment/providers/stripe.ts
```diff
@@ -0,0 +1,227 @@
...
+	stripePublishableKey?: string;
+};
+
+export type StripeGatewayConfig = {
+	id: string;
+	data?: StripeGatewayConfigData | null;
+};
+
+export function isStripeGateway(gatewayId: string): boolean {
+	return gatewayId === STRIPE_GATEWAY_ID || gatewayId.includes("stripe");
```
peelar: **question**: why the fallback here? We have to be sure if we are looking at this specific Stripe payment app because otherwise the storefront <> stripe integration won't work, right?

peelar (dismissed):
didn't investigate too deeply but beware of:

> Warning: Custom Cache-Control headers detected for the following routes:
>   - /_next/static/chunks/:path*
>   - /_next/static/:path*
> 
> Setting a custom Cache-Control header can break Next.js development behavior.
> - Cache Components enabled
> - Experiments (use with caution):
>   · optimizePackageImports

src/checkout/lib/payment/providers/stripe.ts
mirekm: Good catch. `gatewayId.includes("stripe")` is a leftover from the previous iteration.

peelar (approved):

PR #1215, opened 2026-06-26
Document Stripe sandbox setup in README

mirekm:


README.md
```diff
@@ -401,6 +401,55 @@ STOREFRONT_DISCOVER_CHANNELS=true            # Opt-in: discover ALL active Saleo
...
+
+3. **Enable Stripe in the storefront.** Add to `.env` (or `.env.local`):
+
+   ```bash
+   NEXT_PUBLIC_ENABLE_STRIPE_PAYMENTS=true
+   ENABLE_STRIPE_PAYMENTS=true   # server-side mirror for the transaction guard
+
+   # Optional — wallet buttons (Apple Pay / Google Pay / Link); on by default when Stripe is enabled
+   # NEXT_PUBLIC_ENABLE_STRIPE_EXPRESS_CHECKOUT=false
+   ```
```
peelar, line 436: I am skeptical if we need the env vars controls. I think if the app is installed, the payment method should display here.

peelar (approved):

PR #1233, opened 2026-07-29
feat(accounts): add password confirmation for sign-up

NyanKiyoshi:
This adds a new `password` input field when activating the account which implements the upgrade notes from https://docs.saleor.io/upgrade-guides/core/migrate-account-merging

Note: this accepts any password by default, this is normal. As per the upgrade guide: `accountConfirmMergeMode` must be set to `REQUIRE_PASSWORD` for the `password` input to be checked

### DIFF
--- a/src/ui/components/auth/confirm-account-mode.tsx
+++ b/src/ui/components/auth/confirm-account-mode.tsx
@@ -3,14 +3,22 @@
 import { useEffect, useRef, useState } from "react";
 import Link from "next/link";
 import { useTranslations } from "next-intl";
-import { CheckCircle, Loader2, Mail } from "lucide-react";
+import { CheckCircle, Eye, EyeOff, Lock, Mail } from "lucide-react";
 import { confirmAccountWithBffDeduped } from "@/lib/auth/confirm-account-client";
 import { buildStorefrontPath } from "@/lib/storefront-path";
-import { buttonClassName } from "@/ui/components/ui/button";
+import { Button, buttonClassName } from "@/ui/components/ui/button";
+import { Input } from "@/ui/components/ui/input";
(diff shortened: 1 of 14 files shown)


src/ui/components/auth/confirm-account-mode.tsx
```diff
@@ -3,14 +3,22 @@
...
+import { CheckCircle, Eye, EyeOff, Lock, Mail } from "lucide-react";
 import { confirmAccountWithBffDeduped } from "@/lib/auth/confirm-account-client";
 import { buildStorefrontPath } from "@/lib/storefront-path";
-import { buttonClassName } from "@/ui/components/ui/button";
+import { Button, buttonClassName } from "@/ui/components/ui/button";
+import { Input } from "@/ui/components/ui/input";
+import { Label } from "@/ui/components/ui/label";
 
-type ConfirmState = "confirming" | "success" | "error";
+type ConfirmState = "form" | "success" | "error";
```
peelar, line 13: **thought**: `form` seems like a strange name for state enum value but maybe I lack context

peelar (approved):

src/ui/components/auth/confirm-account-mode.tsx
NyanKiyoshi: The `form` state is the state shows the account activation form - what do you suggest as the name instead if seems strange? `confirming` makes less sense IMO as it sounds like a loading/pending state rather than a state where the user is actually doing something

peelar: I read through more code and I think it makes sense, I thought it's a standard "action not started", "action success", "action error" type of state.

